Issue #57, opened 2020-02-10
images


tischi:
<img width="468" alt="image" src="https://user-images.githubusercontent.com/2157566/74152530-c9603500-4c0e-11ea-81d8-d518b9ed1ef3.png">
<img width="851" alt="image" src="https://user-images.githubusercontent.com/2157566/74152904-98cccb00-4c0f-11ea-9819-6c772174a2c0.png">
<img width="249" alt="image" src="https://user-images.githubusercontent.com/2157566/74152986-cc0f5a00-4c0f-11ea-869d-23456fec0cdc.png">
<img width="600" alt="image" src="https://user-images.githubusercontent.com/2157566/74168546-37febc00-4c2a-11ea-9981-85232ed5322e.png">
<img width="220" alt="image" src="https://user-images.githubusercontent.com/2157566/74252332-d2740380-4ced-11ea-9a4e-d25cb1e83859.png">

constantinpape:
Images were moved to `images` folder.

PR #74, opened 2020-02-25
Move Images to folder; update README; remove log file

constantinpape:
- Moves images from #57 to folder `images`
- Updates the README to reflect new UI and use moved images
- remove file `hs_err_pid55005.log`

@tischi to me it looked like the log file was added here by accident. Let me know if there is a reason for it being in the repo, then I will restore it.


tischi:
I trust you 👍

Issue #133, opened 2020-10-11
Bookmark settings persist in source metadata


K-Meech:
Viewing a bookmark currently modifies the sources' metadata - this means that these settings persist in the current session. e.g. If you open the platybrowser, view the bookmark for figure 5C, then remove the cell segmentation from the viewer. Every time you add the cell segmentation after that it will add the additional tables from the bookmark /  colour by column etc..

@tischi - could we possibly separate the default metadata for a source (i.e. from images.json), from the current metadata? This would stop the bookmark settings from persisting, and make it easier to save the current settings to bookmarks. e.g. we could add another Metadata entry into the SourceAndMetadata class?

tischi:
Thank for spotting this! I would almost call it a bug.

It is a bit tricky though, because maybe some modified metadata we want to persist, e.g. if a user change the brightness and contrast display settings.

But for sure I would agree that the metadata changes from a bookmark should not persist.

In fact, the whole `Metadata` class is a bit of a mess, grown historically.

Within MoBIE I tried to improve the situation a bit with those classes: `ImageProperties` and `MutableImageProperties`. But since the imagej-utils code has legacy issues (I also use it in other projects) I had to still use the `Metadata` class in the end and I think for now we have to live with it.

Would the following work?

In `Bookmarksmanager` there is this function, where I am proposing some changes

```
	public void addSourcesToPanelAndViewer( Bookmark bookmark )
	{
		for ( Map.Entry< String, MutableImageProperties> newSettings : bookmark.layers.entrySet() )
		{
			final String sourceName = entry.getKey();
			if ( ! sourcesPanel.getVisibleSourceNames().contains( sourceName ) )
			{
                                // change metdata to reflect what is demanded by the boomark and remember the old settings
				MutableImageProperties oldSettings =  updateSourceMetadataAndFetchPerviousSettings( newSettings, sourceName );
                                // show the source
				sourcesPanel.addSourceToPanelAndViewer( metadata.displayName );
                                /// ??? reset the metadata to what was before
                                updateSourceMetadataAndFetchPerviousSettings( oldSettings, sourceName );
			}
		}
	}
```

Something like this? Do you think that could work?

K-Meech:
Ok - makes sense! This would work.
[...]

Issue #83, opened 2020-05-15
Support making bookmarks from UI


constantinpape:

tischi:
@constantinpape 

Well, thanks to our recent efforts we know now how to submit files to github via the API 😃  
One option would thus be that we have a **bookmarks folder** and each bookmark would be one json file. That would make it quite simple to add and remove bookmarks. I would find that easier and cleaner than changing the one big bookmarks file that we currently have.

What do you think? 

It is a bit annoying though that a user would need to get a github API token and also push rights to the repo. It is quite involved to get this and thus the user would need to be kind of a github crack...speaking of which, the other repo is on gitlab, not github...maybe annoying to even support both github and gitlab API...then again gitlab API looks at first glance very similar to github: https://docs.gitlab.com/ee/api/repository_files.html

constantinpape:
> Well, thanks to our recent efforts we know now how to submit files to github via the API smiley
> One option would thus be that we have a **bookmarks folder** and each bookmark would be one json file. That would make it quite simple to add and remove bookmarks. I would find that easier and cleaner than changing the one big bookmarks file that we currently have.

Yes, I agree that's a better solution then having the big bookmark file.
(We could still support multiple bookmarks per file though).

> It is a bit annoying though that a user would need to get a github API token and also push rights to the repo. It is quite involved to get this and thus the user would need to be kind of a github crack...speaking of which, the other repo is on gitlab, not github...maybe annoying to even support both github and gitlab API...then again gitlab API looks at first glance very similar to github: https://docs.gitlab.com/ee/api/repository_files.html

For starters, we can just support the bookmarks locally. I.e. you have a second folder in the fiji plugin sources (custom data in there is possible, right?) where the bookmarks generated from the UI are stored.
If users want to push those upstream, they need to clone the repo and make a PR with their new bookmarks (this means they also don't need push rights, which is not a scalable solution).

Of course in the long run also being able to push to github / gitlab directly would be great, but I don't think it's immediately necessary to be useful.

tischi:
> Of course in the long run also being able to push to github / gitlab directly would be great

Actually, maybe it is least work to just implement this. I have most of the code in the PlateViewer already anyway.

constantinpape:
> > 
> 
> Actually, maybe it is least work to just implement this. I have most of the code in the PlateViewer already anyway.

Ok, but from a user perspective there should always be a local option to store the bookmarks,
so that it's not necessary to have git API tokens.

[...]

PR #134, opened 2020-10-14
Bookmark ui

K-Meech:
Main changes:
- Only bookmarks in default.json are loaded by default
- Other bookmarks are loaded via the context menu 'load additional bookmarks..' and follows the same logic (and uses most of the same code) as loading additional tables. i.e. can select from file on project or file system
- Bookmarks can be saved via the context menu 'save current settings as bookmark..' with similar logic to above. i.e. can select to save to project or to file system. For saving to a project on github, at the moment this works the same as PlateViewer, and requires an access token with push access to the repo. For saving to project or file system, if you select a file that already exists, it will append to that file. (appending to json files is actually a pain, so really it parses the file, appends the new bookmark and saves it again - it works ok though)
- Bookmark settings no longer persist

There are also changes to imagej-utils - pull request soon!

Deals with issues: https://github.com/mobie/mobie-viewer-fiji/issues/133 and https://github.com/mobie/mobie-viewer-fiji/issues/83

### DIFF
--- a/src/main/java/de/embl/cba/mobie/bookmark/BookmarksJsonParser.java
+++ b/src/main/java/de/embl/cba/mobie/bookmark/BookmarksJsonParser.java
@@ -48,8 +48,13 @@ public class BookmarksJsonParser {
 		try {
 			ArrayList<String> filePaths = new ArrayList<>();
 			String bookmarksDirectory = FileAndUrlUtils.combinePath(datasetLocation, "misc", "bookmarks");
-			filePaths.add( selectPathFromProjectOrFileSystem( bookmarksDirectory, "Bookmark" ) );
-			return parseBookmarks(filePaths);
+			String selectedFilePath = selectPathFromProjectOrFileSystem( bookmarksDirectory, "Bookmark" );
+			if (selectedFilePath != null) {
+				filePaths.add(selectedFilePath);
+				return parseBookmarks(filePaths);
+			} else {
+				return null;
+			}
 		} catch (IOException e) {
 			e.printStackTrace();
 			return null;
--- a/src/main/java/de/embl/cba/mobie/bookmark/BookmarksJsonParser.java
+++ b/src/main/java/de/embl/cba/mobie/bookmark/BookmarksJsonParser.java
@@ -144,32 +144,50 @@ public class BookmarksJsonParser
 		final JFileChooser jFileChooser = new JFileChooser();
 		jFileChooser.setFileFilter(new FileNameExtensionFilter("json", "json"));
 		if (jFileChooser.showSaveDialog(null) == JFileChooser.APPROVE_OPTION) {
-			jsonPath = jFileChooser.getSelectedFile().getAbsolutePath() + ".json";
+			jsonPath = jFileChooser.getSelectedFile().getAbsolutePath();
 		}
 
 		if (jsonPath != null) {
 
-			// exclude the name field from json
-			ExclusionStrategy strategy = new ExclusionStrategy() {
-				@Override
-				public boolean shouldSkipField(FieldAttributes f) {
-					if (f.getName().equals("name")) {
-						return true;
+			if (!jsonPath.endsWith(".json")) {
+				jsonPath += ".json";
+			}
+
+			File jsonFile = new File(jsonPath);
+			if (jsonFile.exists()) {
+				int result = JOptionPane.showConfirmDialog(null,
+						"This Json file already exists - append bookmark to this file?", "Append to file?",
+						JOptionPane.YES_NO_OPTION,
+						JOptionPane.QUESTION_MESSAGE);
+				if (result != JOptionPane.YES_OPTION) {
+					jsonFile = null;
+				}
+			}
+
+			if (jsonFile != null) {
+
+				// exclude the name field from json
+				ExclusionStrategy strategy = new ExclusionStrategy() {
+					@Override
+					public boolean shouldSkipField(FieldAttributes f) {
+						if (f.getName().equals("name")) {
+							return true;
+						}
+						return false;
 					}
-					return false;
-				}
 
-				@Override
-				public boolean shouldSkipClass(Class<?> clazz) {
-					return false;
-				}
-			};
+					@Override
+					public boolean shouldSkipClass(Class<?> clazz) {
+						return false;
+					}
+				};
 
-			Gson gson = new GsonBuilder().addSerializationExclusionStrategy(strategy).create();
-			Type type = new TypeToken<Map<String, Bookmark>>() {
-			}.getType();
+				Gson gson = new GsonBuilder().addSerializationExclusionStrategy(strategy).create();
+				Type type = new TypeToken<Map<String, Bookmark>>() {
+				}.getType();
 
-			writeBookmarksToFile(gson, type, jsonPath, namesToBookmarks);
+				writeBookmarksToFile(gson, type, jsonFile, namesToBookmarks);
+			}
 		}
 
 	}
@@ -184,9 +202,16 @@ public class BookmarksJsonParser
 		return stringBookmarkMap;
 	}
 
-	private void writeBookmarksToFile (Gson gson, Type type, String filePath, Map< String, Bookmark > bookmarks) throws IOException
+	private void writeBookmarksToFile (Gson gson, Type type, File jsonFile, Map< String, Bookmark > bookmarks) throws IOException
 	{
-		OutputStream outputStream = new FileOutputStream( new File( filePath ) );
+		Map<String, Bookmark> bookmarksInFile = new HashMap<>();
+		// If json already exists, read existing bookmarks to append new ones
+		if (jsonFile.exists()) {
+			bookmarksInFile.putAll(readBookmarksFromFile(gson, type, jsonFile.getAbsolutePath()));
+		}
+		bookmarksInFile.putAll(bookmarks);
+
+		OutputStream outputStream = new FileOutputStream( jsonFile );
 		final JsonWriter writer = new JsonWriter( new OutputStreamWriter(outputStream, "UTF-8"));
 		writer.setIndent("	");
 		gson.toJson(bookmarks, type, writer);
--- a/src/main/java/de/embl/cba/mobie/bookmark/BookmarksJsonParser.java
+++ b/src/main/java/de/embl/cba/mobie/bookmark/BookmarksJsonParser.java
@@ -4,18 +4,10 @@ import com.google.gson.ExclusionStrategy;
 import com.google.gson.FieldAttributes;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
-import com.google.gson.internal.LinkedTreeMap;
 import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
-import de.embl.cba.tables.github.GitHubUtils;
-import de.embl.cba.tables.github.GitLocation;
-import de.embl.cba.mobie.image.SourcesModel;
 import de.embl.cba.tables.FileAndUrlUtils;
-import de.embl.cba.tables.FileUtils;
-import de.embl.cba.tables.github.GitHubContentGetter;
-import de.embl.cba.tables.github.GitHubUtils;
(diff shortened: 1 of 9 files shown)


src/main/java/de/embl/cba/mobie/bookmark/BookmarksJsonParser.java
```diff
@@ -135,6 +90,96 @@ public BookmarksJsonParser( String datasetLocation )
...
 
+	public void saveBookmarksToGithub(ArrayList<Bookmark> bookmarks) {
+		final GitLocation gitLocation = GitHubUtils.rawUrlToGitLocation( datasetLocation );
+		gitLocation.path += "misc/bookmarks";
+
+		BookmarkGithubWriter bookmarkWriter = new BookmarkGithubWriter(gitLocation, this);
+		bookmarkWriter.writeBookmarksToGithub(bookmarks);
+	}
+
+	private Gson createBookmarkWriterGson (boolean usePrettyPrinting) {
```
tischi, line 101: Maybe rename to
`createGsonBuilder`
?!

src/main/java/de/embl/cba/mobie/bookmark/BookmarksManager.java
```diff
@@ -75,6 +91,131 @@ public void adaptViewerTransform( Bookmark bookmark )
...
+		String bookmarkName = null;
+		final GenericDialog gd = new GenericDialog( "Choose save location" );
+		gd.addStringField("Bookmark Name", "name");
+		gd.addChoice( "Save to", new String[]{FileUtils.PROJECT, FileUtils.FILE_SYSTEM }, FileUtils.PROJECT );
+		gd.showDialog();
+		if ( gd.wasCanceled() ) return null;
+		bookmarkName = gd.getNextString();
+		fileLocation = gd.getNextChoice();
+
+		ArrayList<String> bookmarkNameandLocation = new ArrayList<>();
```
tischi, line 134: This is confusing, because a `List` could have many items.

Just make a little non public class within the `BookmarksManger`

```
class NameAndLocation {
  String name;
  String location;
}
```

src/main/java/de/embl/cba/mobie/bookmark/BookmarksManager.java
```diff
@@ -75,6 +91,131 @@ public void adaptViewerTransform( Bookmark bookmark )
...
+		bookmarkNameandLocation.add(bookmarkName);
+		bookmarkNameandLocation.add(fileLocation);
+		return bookmarkNameandLocation;
+	}
+
+	public Bookmark getBookmarkFromCurrentSettings( String bookmarkName) {
```
tischi, line 140: My personal convention is (and there is no general convention):

- If a method `creates` something new I call it `createXYZ( )`
- If a method just returns something that exists already I call it `getXYZ( )`

In this case I would thus vote for calling it:

`createBookmarkFromCurrentSettings( String newBookmarkName )`

Otherwise I find it confusing, the way it is right now it sounds more like it tries to fetch a bookmark with the name `bookmarkName` from an already existing list of bookmarks.

src/main/java/de/embl/cba/mobie/bookmark/BookmarksManager.java
```diff
@@ -75,6 +91,131 @@ public void adaptViewerTransform( Bookmark bookmark )
...
+		double[] currentPosition = new double[3];
+		BdvUtils.getGlobalMouseCoordinates(bdv).localize(currentPosition);
+		currentBookmark.position = currentPosition;
+		currentBookmark.normView = Utils.createNormalisedViewerTransformString( bdv, Utils.getMousePosition( bdv ) ).split(",");
+		currentBookmark.view = null;
+
+		return currentBookmark;
+	}
+
+	private MutableImageProperties getCurrentImageProperties(String sourceName) {
```
tischi, line 162: I would rename to 

`fetchMutableSourceProperties( String sourceName )`

`fetch` being different from `get` and `create` in a sense that it does no really `create` something new but just looks around and `fetches` stuff ;-)
but whatever... :-)

src/main/java/de/embl/cba/mobie/image/ImagePropertiesToMetadataAdapter.java
```diff
@@ -32,4 +34,21 @@ public void setMetadata( Metadata metadata, MutableImageProperties imageProperti
 		metadata.showSelectedSegmentsIn3d = imageProperties.showSelectedSegmentsIn3d;
 		metadata.additionalSegmentTableNames = imageProperties.tables;
 	}
+
+	public void setImageProperties( Metadata metadata, MutableImageProperties imageProperties) {
```
tischi, line 38: rename to `setMutableImagePropertiesFromMetadata` and change the order of the arguments:

`setMutableImagePropertiesFromMetadata( MutableImageProperties properties, Metadata metadata )`

My above method `setMetadata` should be renamed accordingly:

`setMetadataFromMutableImageProperties`

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesDisplayUI.java
```diff
@@ -89,7 +89,7 @@ public static JButton createBrightnessButton( int[] buttonDimensions,
 	}
 
 
-	private static ArrayList< ConverterSetup > getConverterSetups(
+	public static ArrayList< ConverterSetup > getConverterSetups(
```
tischi, line 92: Since it is `public static` now maybe move to some `Utils` class?

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
```diff
@@ -120,8 +121,6 @@ private void addColorButton( JPanel panel, int[] buttonDimensions, SourceAndMeta
 
     private void setSourceColor( SourceAndMetadata< ? > sam, Color color, JPanel panel )
     {
-        sam.metadata().color = color.toString();
```
tischi, line 123: Why don't you change this field anymore?

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
```diff
@@ -146,7 +145,7 @@ public void setSourceColor( String sourceName, Color color )
             return;
         }
 
-        final SourceAndMetadata< ? > sam = getSourceAndMetadata( sourceName );
+        final SourceAndMetadata< ? > sam = sourceNameToSourceAndMetadata.get(sourceName);
```
tischi, line 148: Here you directly access:

`sourceNameToSourceAndMetadata.get(...)`

and below...

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
```diff
@@ -291,17 +290,21 @@ public void addSourceToPanelAndViewer( String sourceName )
         }
         else
         {
-            final SourceAndMetadata< ? > sam = getSourceAndMetadata( sourceName );
-            sourceNameToSourceAndMetadata.put( sam.metadata().displayName, sam );
+            final SourceAndMetadata< ? > sam = getSourceAndDefaultMetadata( sourceName );
```
tischi, line 293: ...and here you go through some getter method:

`getSourceAndDefaultMetadata(...`

Is there a reason? If not, maybe do the same thing in both cases?

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
```diff
@@ -541,6 +545,7 @@ private void mergeDefaultTableWithAdditionalTables( SourceAndMetadata< ? > sam,
         for ( String tableName : sam.metadata().additionalSegmentTableNames )
         {
             String newTablePath = FileAndUrlUtils.combinePath( tablesLocation, tableName + ".csv" );
+            tableRowsTableView.addAdditionalTables(newTablePath);
```
tischi, line 548: Just a general comment. In `imglib2` the convention is to have a lot of spaces.

For example:

`tableRowsTableView.addAdditionalTables( newTablePath );`

instead of

`tableRowsTableView.addAdditionalTables(newTablePath);`

but "normal" Java would be without the spaces.

`MoBIE` tries to adhere to `imglib2` style. So, if you happen to think about it, it would be nice if you added the spaces. But don't worry now. Maybe one can just tell IntelliJ to make all of this consistent at some point.

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
```diff
@@ -706,13 +711,14 @@ public void removeSourceFromPanelAndViewers( String sourceName )
...
 
-    public void removeSourceFromPanelAndViewers( SourceAndMetadata< ? > sam )
+    private void removeSourceFromPanelAndViewers( SourceAndMetadata< ? > sam )
     {
         updateSegments3dView( sam.metadata(), this, false );
         updateSource3dView( sam, this, false );
 
         removeSourceFromPanel( sam.metadata().displayName );
 		removeLabelViews( sam.metadata().displayName );
+		sourceNameToSourceAndMetadata.remove( sam.metadata().displayName );
```
tischi, line 721: Seeing this now, maybe we should rename for sake of clarity:
`sourceNameToSourceAndCurrentMetadata`
to make clear that this is something temporary that can be removed?

Other question:
If the user comes back and adds the source a second time to the viewer, which display settings would be applied?
I guess the default ones? Is that what we want? We can leave it like this for now for simplicity.

tischi (commented):
Again: Amaaaaaazing!
You rock!

Just few comments...

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
K-Meech: I changed this because the output from color.toString() can't be read when this metadata is loaded again. Currently, this means that if you add an image (say one of the prospr genes), change the colour, remove it, and then re-add it again, it will appear white. This is why I added support for a "(r=255,g=255,b=255,a=255)" format in imagej-utils, and this is the version I save to the bookmarks. This does mean that colours can't persist though...

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
K-Meech: This is due to the default metadata vs current metadata. In the setColor call it needs to get the current metadata (i.e. maybe the bookmark metadata that was used to create it). In the addSourceToPanelAndViewer, it needs to get the actual default data so that bookmarks don't persist etc... I'll rename sourceNameToSourceAndMetadata to make this clearer

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
K-Meech: Ok! Will bear this in mind for the future

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
K-Meech: Yes - for the moment, this will apply the default settings, so nothing persists. I guess later, if there are certain things we'd like to persist e.g. contrast settings etc..., then we can make these modify the default metadata directly? The issue with using the current metadata is that everything would persist > additional tables / colouring by columns etc which I'm not sure is what we want..

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesPanel.java
tischi: I think you have to explain me in person... :-)

src/main/java/de/embl/cba/mobie/ui/viewer/SourcesDisplayUI.java
K-Meech: I moved this to BdvUtils in imagej-utils

tischi: Cool!

K-Meech:
All changes made! (and on the imagej-utils request)

Issue #135, opened 2020-10-28
Enums....


tischi:
@K-Meech 

I did not find the issue, but I think we considered making something an `enum` 🙂 

I found some code that shows how to use them in a dialog:

```

public enum Level
{
	Normal,
	Debug,
	Benchmark
}

public synchronized static void showLoggingLevelDialog()
	{
		final GenericDialog gd = new GenericDialog( "Logging" );
		gd.addChoice( "Level", new String[]{
				Level.Debug.toString(), Level.Normal.toString(), Level.Debug.toString()
		}, Level.Normal.toString() );
		gd.addMessage( "Benchmark mode may significantly slow down the application!" );

		gd.showDialog();
		if ( gd.wasCanceled() ) return;
		setLevel( Level.valueOf( gd.getNextChoice() ) );
	}
```

K-Meech:
Thanks! I made the changes - it was the enum for Project vs File_system location that needed changing.

PR #136, opened 2020-10-28
Switch to enum for file location

K-Meech:
Match to changes in imagej-utils - uses enum for file location.
For https://github.com/mobie/mobie-viewer-fiji/issues/135

### DIFF
--- a/src/main/java/de/embl/cba/mobie/bookmark/BookmarksManager.java
+++ b/src/main/java/de/embl/cba/mobie/bookmark/BookmarksManager.java
@@ -108,7 +108,7 @@ public class BookmarksManager
 		ArrayList<Bookmark> bookmarks = new ArrayList<>();
 		bookmarks.add(currentBookmark);
 
-		if ( bookmarkNameAndLocation.location == FileLocation.PROJECT &&
+		if ( bookmarkNameAndLocation.location == FileLocation.Project &&
 				bookmarksJsonParser.getDatasetLocation().contains( "raw.githubusercontent" )) {
 			bookmarksJsonParser.saveBookmarksToGithub(bookmarks);
 		} else {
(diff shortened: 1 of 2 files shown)


src/main/java/de/embl/cba/mobie/bookmark/BookmarksManager.java
```diff
@@ -108,7 +108,7 @@ public void saveCurrentSettingsAsBookmark () {
 		ArrayList<Bookmark> bookmarks = new ArrayList<>();
 		bookmarks.add(currentBookmark);
 
-		if (bookmarkNameAndLocation.location.equals(FileUtils.PROJECT) &&
+		if ( bookmarkNameAndLocation.location == FileLocation.Project &&
```
tischi, line 111: I think it is better practice to use `.equals( )` instead of `==`.
For example for Strings `==` yields unexpected behaviour.

src/main/java/de/embl/cba/mobie/bookmark/BookmarksManager.java
```diff
@@ -122,19 +122,26 @@ public void saveCurrentSettingsAsBookmark () {
...
 		gd.addStringField("Bookmark Name", "name");
-		gd.addChoice( "Save to", new String[]{FileUtils.PROJECT, FileUtils.FILE_SYSTEM }, FileUtils.PROJECT );
+		gd.addChoice( "Save to", new String[]{ FileLocation.Project.toString(),
+				FileLocation.File_system.toString() }, FileLocation.Project.toString() );
 		gd.showDialog();
+
 		if ( gd.wasCanceled() ) return null;
 		bookmarkName = gd.getNextString();
-		fileLocation = gd.getNextChoice();
+		String fileLocationString = gd.getNextChoice();
```
tischi, line 139: also here, use the `FileLocation.valueOf()`

tischi (commented):
....some suggestions.

K-Meech:
Fixed!

PR #143, opened 2020-11-09
add resolution 3d

K-Meech:
I added the 3d resolution to the metadata / imageproperties. 
I will need to add a bit to the function here: https://github.com/mobie/mobie-viewer-fiji/blob/master/src/main/java/de/embl/cba/mobie/bookmark/BookmarksManager.java#L170 when you're done. It needs a getter for the current resolution 3d value. This is so the current value can be written to a bookmark, if it has been changed via the context menu.


tischi:
> It needs a getter for the current resolution 3d value. 

I see what you mean, but the question (I don't know the answer myself) is _from whom_ to get it....
Would it be OK, if I just change the `currentMetadata` and you read it from there?

Anyway, THANKS!

K-Meech:
Yes - that should work!

PR #147, opened 2020-11-20
WIP Open S3 Zarr in MoBIE

tischi:
@joshmoore
I created this branch for development.

Milestone:

- [ ] Get this function working: https://github.com/mobie/mobie-viewer-fiji/blob/s3zarr/src/test/java/develop/s3/OpenN5S3ZarrXML.java

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
 	</parent>
 
 	<artifactId>mobie</artifactId>
-	<version>0.5.3</version>
+	<version>0.5.4-SNAPSHOT</version>
 
 	<url>https://github.com/mobie/mobie-viewer-fiji</url>
 	<name>MoBIE Fiji Viewer</name>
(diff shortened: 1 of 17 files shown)


jburel:
@tischi do we need that specific snapshot for that current Zarr work ``<j3dcore.version>1.6.0-scijava-3-SNAPSHOT</j3dcore.version> ``. I am preparing some docs and ideally it will be good if we could reduce what one has to build locally.

tischi:
I changed it to the other non SNAPSHOT version `1.6.0-scijava-2` as for the ome.zarr stuff it  does not matter, it is just for other MoBIE functionality.

For our course end of the month I would anyway factor out the ome.zarr stuff into a separate repo and then this dependency would be gone completely: https://github.com/mobie/mobie-viewer-fiji/issues/161

jburel:
Great. Thanks. I will look for the new repo so I can add it to the guide.

PR #163, opened 2020-12-16
replace spaces in names

K-Meech:
Replaces spaces in all input with _


tischi:
Cool. 

I would rename `tidyStringInput` to just `tidyString`.

If you are at it, maybe ensure that there are only alphanumerics and `_`?!

Some people put amazing chars into their names, including: `/ \ , ! . ; ` all of which are very likely to break something somewhere.

I think you could also have something like:

```
if ( name != tidyString(name ) )
{
     \\ some feedback to the user
}
```

K-Meech:
Good point - I'll add that in!

K-Meech:
I added checks that it's only alphanumerics, _, or -. Also it gives some feedback in the log for any changes.

PR #168, opened 2021-01-09
build fix

jburel:
I had to build from source in order to test the ``OMEZarrS3Reader`` (using the beta site did not have the file)
I had to add the following dependency to the pom in order for it to work and upgrade my maven version (change in readme)

cc @tischi


tischi:
@jburel 
Thanks a lot!

PR #175, opened 2021-01-28
Affine support

K-Meech:
Adds a field for an affine transform to the 'add current image' dialog for mobie project creator.


K-Meech:
@tischi Could you merge this and put it on the main update site? I need it for the demo on Wednesday. Thanks!

tischi:
Should be done now.

PR #184, opened 2021-02-04
Improved N5 Export

tischi:
@K-Meech 

I'll try here to make a list of things we want to achieve:

- Create a class that can save an ImagePlus to N5, based on the ExportImagePlusAsN5PlugIn in bdv-fiji
  - We would copy and paste the code into this repo and modify it to 
    - separate UI from functionality
    - add the option to specify a custom affine transform
    - add the option to chose the downsampling modality (avg, nearest)
- We copy WriteSequenceToN5 from bdv-core and modify it such that the downsampling modality can be configured

In general we try to make the changes such that our work could be merged at some point into bdv-fiji and bdv-core.

What do you think?

### DIFF
--- a/src/main/java/de/embl/cba/mobie/n5/export/WriteSequenceToN5.java
+++ b/src/main/java/de/embl/cba/mobie/n5/export/WriteSequenceToN5.java
@@ -1,0 +1,355 @@
+/*-
+ * #%L
+ * BigDataViewer core classes with minimal dependencies.
+ * %%
+ * Copyright (C) 2012 - 2020 BigDataViewer developers.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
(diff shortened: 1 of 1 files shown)


K-Meech:
Sounds good to me!
I'm working on ExportImagePlusAsN5Plugin at the moment, so if you could look at WriteSequencetoN5 - that'd be perfect!
Ideally we'd also make these changes for ExportImagePlusPlugin https://github.com/bigdataviewer/bigdataviewer_fiji/blob/master/src/main/java/bdv/ij/ExportImagePlusPlugIn.java and WriteSequenceToHdf5 https://github.com/bigdataviewer/bigdataviewer-core/blob/master/src/main/java/bdv/export/WriteSequenceToHdf5.java. It's not so urgent, but it'd be nice to still support hdf5. As far as I can see from the code, it is structured in an extremely similar way, so shouldn't be too hard to do.

PR #203, opened 2021-03-02
BDV 10.0, major refactoring and support of image grids

tischi:
@K-Meech 
This does not use the BDV cards UI, but contains major refactoring to make that easier, if needed.
Could you have a look whether we could merge this already now?


K-Meech:
@tischi I tried it out locally, and I noticed that the context menus are gone.
e.g. the 'load additional bookmarks...' option. Only the options for displayed segmentations are still there.

tischi:
Does it work now?

K-Meech:
That fixed the context menu!
I found another though - when you view a bookmark (e.g. Figure2C:Muscle Segmentation for the platy), it adds the raw data twice.

tischi:
Good find!
Would you have time to look for the bug and then commit the fix to this branch?
Like this you will also have a chance to look at bit at the new structure of the code (still not perfect, but I hope a bit cleaner).

K-Meech:
Sure - makes sense! I'll have some time tomorrow to look through

K-Meech:
@tischi I fixed the bug with the bookmarks, and pushed it to this branch

tischi:
Great!  Could we use this branch as our sort-of-master now for further developments?
Maybe until the whole grid view stuff is working including an image table?

K-Meech:
Sure - sounds good!

tischi:
@K-Meech I think it is working now. Could you try? There is this class in test/java/projects: OpenRemoteTomogramsBranch

K-Meech:
It's not building for me - says there's no 'misc' in Metadata. Did you make some local changes to Metadata in imagej-utils?

tischi:
Sorry, forgot to push changes to imagej-utils. Now it should be there.

K-Meech:
Ok  - I see it now! There seems to be a problem with the colouring scheme though. e.g. Here I clicked on the top row of the table: it went to the correct position, but the tomogram is coloured purple, and not the green shown in the table.
Screenshot (261)

tischi:
@K-Meech I think I fixed the bug.

K-Meech:
Yes - it all works now!

Issue #209, opened 2021-03-10
Can't open Platybrowser on current master


constantinpape:
When trying to open the platybrowser on the current master, it fails with the following error:

```
Adding source: sbem-6dpf-1-whole-raw...
Exception in thread "AWT-EventQueue-1" java.lang.NoClassDefFoundError: bdv/BehaviourTransformEventHandler3D
	at bdv.util.BdvOptions$Values.<init>(BdvOptions.java:277)
	at bdv.util.BdvOptions.<init>(BdvOptions.java:51)
	at bdv.util.BdvOptions.options(BdvOptions.java:59)
	at de.embl.cba.mobie.ui.SourcesDisplayManager.addSourceToBDV(SourcesDisplayManager.java:459)
	at de.embl.cba.mobie.ui.SourcesDisplayManager.showIntensitySource(SourcesDisplayManager.java:444)
	at de.embl.cba.mobie.ui.SourcesDisplayManager.show(SourcesDisplayManager.java:415)
	at de.embl.cba.mobie.bookmark.BookmarkManager.show(BookmarkManager.java:93)
	at de.embl.cba.mobie.bookmark.BookmarkManager.setView(BookmarkManager.java:58)
	at de.embl.cba.mobie.ui.MoBIE.lambda$new$0(MoBIE.java:78)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at org.GNOME.Accessibility.AtkWrapper$6.dispatchEvent(AtkWrapper.java:705)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
[...]

Issue #182, opened 2021-02-02
Support loading a project directly from s3


constantinpape:
Enable opening projects from s3 directly, without the need for github.
This would make it easier to share data for users who are not familiar with git/github.

Having the metadata on github should always stay an option for legacy with old projects and to have version control for the metadata.

constantinpape:
@tischi I can maybe look into this, but I would need some pointers to the relevant parts in the code.

constantinpape:
@tischi quick reminder to point me to the relevant part of the viewer code.

tischi:
@constantinpape

https://github.com/mobie/mobie-viewer-fiji/blob/1c4f8d296fe505f4f421ef101b428ba222db0480/src/main/java/de/embl/cba/mobie/ui/MoBIE.java#L60

constantinpape:
I have implemented access for public buckets in #211.

Now I am looking into implementing access from private buckets (i.e. buckets that need credentials for read access).
While working on this I noticed a few things and ran into some issues:
1. It's actually possible to access files in public buckets via HTTPS (I initially thought this wasn't possible), but we still need special treatment for s3 in order to list objects etc.
2. It's not necessary to know beforehand if a bucket is public or private, so we could get rid of the "Protected" and "Anonymous" fields in the xml. I don't actually know how to do this in java yet, only in python, I will have a look into how to do it in java and then refactor the code a bit.
3. Accessing files in private buckets via HTTPS doesn't work. This currently prevents us from opening a project from a private buket, because the bdv-xmls are still streamed via HTTPS. I am not quite sure what's the best way to fix this, because it looks to me like this is happening in `mpicbg.spim.data.generic.XmlIoAbstractSpimData.load` so not in a place we can easily change. See full stacktrace below for details.

@tischi I think it's best if we discuss this via zoom or phone (I have time all Monday).

For reference, here is the full traceback when trying to open from a private bucket via HTTPS.
```
Adding source: em-raw...
https://s3.embl.de/platybrowser-private/example-project/example-dataset/images/remote/em-raw.xml
mpicbg.spim.data.SpimDataIOException: java.io.IOException: Server returned HTTP response code: 403 for URL: https://s3.embl.de/platybrowser-private/example-project/example-dataset/images/remote/em-raw.xml
	at mpicbg.spim.data.generic.XmlIoAbstractSpimData.load(XmlIoAbstractSpimData.java:88)
[...]

PR #211, opened 2021-03-10
Open projects from s3

constantinpape:
Towards opening projects directly from s3, see #182.

First change: allows opening projects without the "data" subfolder. This is important for s3 projects, where we don't always follow this convention; and generally nice to have.
I couldn't fully test it due to #209, but I think this part should work.

The next steps toward opening s3 projects directly is using the s3 api for fetching the `datasets.json` and other files (pure http access is not supported in s3, or a least not trivially).

@tischi:
- I think the `fileOrUrlExists` function is pretty useful; do you have something like this implemented somewhere else already? If not, should it be refactored, e.g. into FileAndUrlUtils?
- What's the best strategy for integrating the s3 api (https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/examples-s3.html)? Should this also go int FileAndUrlUtils? Somewhere else?


constantinpape:
> Wow! @constantinpape writing Java code! This is amazing!!
> :-)

It's just weird C++ syntax ;).



> Let's also chat about these things today.

Sounds good.

constantinpape:
This works now.
mobie-from-s3
It depends on the changes in https://github.com/tischi/imagej-utils/pull/36

constantinpape:
Now also opening tables etc. works and I think this is ready to be merged once https://github.com/tischi/imagej-utils/pull/36 gets merged.
The only thing not supported yet is reading from buckets with credentials, see https://github.com/tischi/imagej-utils/pull/36#issuecomment-797506749 for details.

tischi:
Cool! Thanks so much!

PR #269, opened 2021-05-03
make copies of imageDisplay and segmentationDisplay

K-Meech:
Makes serialisable copes of imageDisplay and segmentationDisplay

@tischi I wasn't sure where to fetch ```showImagesIn3d``` for the imageDisplay. Showing images in 3D isn't supported in the viewer yet, so I'm not sure where this info will be.

### DIFF
--- a/src/main/java/de/embl/cba/mobie2/display/SegmentationDisplay.java
+++ b/src/main/java/de/embl/cba/mobie2/display/SegmentationDisplay.java
@@ -99,7 +99,7 @@ public class SegmentationDisplay extends Display
 			this.opacity = ( ( AdjustableOpacityColorConverter ) sourceAndConverter.getConverter() ).getOpacity();
 		}
 
-		this.lut = segmentationDisplay.tableViewer.getColoringLUTName();
+		this.lut = segmentationDisplay.coloringModel.getColoringLUTName();
 		this.colorByColumn = segmentationDisplay.tableViewer.getColoringColumnName();
 
 		Double[] valueLimits = new Double[2];
@@ -119,7 +119,7 @@ public class SegmentationDisplay extends Display
 		}
 
 		this.showSelectedSegmentsIn3d = segmentationDisplay.segmentsVolumeViewer.getShowSegments();
-		this.showScatterPlot = segmentationDisplay.scatterPlotViewer.getShowScatterPlot();
+		this.showScatterPlot = segmentationDisplay.scatterPlotViewer.isVisible();
 		this.scatterPlotAxes = segmentationDisplay.scatterPlotViewer.getSelectedColumns();
 		this.tables = segmentationDisplay.tableViewer.getAdditionalTables();
 	}
--- a/src/main/java/de/embl/cba/mobie2/display/SegmentationDisplay.java
+++ b/src/main/java/de/embl/cba/mobie2/display/SegmentationDisplay.java
@@ -10,6 +10,7 @@ import de.embl.cba.mobie2.bdv.SegmentationImageSliceView;
 import de.embl.cba.mobie2.volume.SegmentsVolumeView;
 import de.embl.cba.mobie2.table.TableViewer;
 import de.embl.cba.tables.color.ColoringLuts;
+import de.embl.cba.tables.color.ColumnColoringModel;
 import de.embl.cba.tables.select.SelectionModel;
 import de.embl.cba.tables.tablerow.TableRowImageSegment;
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


src/main/java/de/embl/cba/mobie2/plot/ScatterPlotViewer.java
```diff
@@ -363,6 +368,7 @@ public void hide()
 	{
 		if ( window != null )
 			window.setVisible( false );
+		showScatterPlot = false;
```
tischi, line 371: @K-Meech I think we don't need the `showScatterPlot`, we could just do `((window != null) && window.isVisible())` to check for visibility. 
And instead of `getShowScatterPlot` I would rename to `isVisible()`

src/main/java/de/embl/cba/mobie2/display/SegmentationDisplay.java
```diff
@@ -75,7 +78,50 @@ public boolean showScatterPlot()
...
+		}
+
+		final SourceAndConverter< ? > sourceAndConverter = segmentationDisplay.sourceAndConverters.get( 0 );
+
+		if( sourceAndConverter.getConverter() instanceof AdjustableOpacityColorConverter)
+		{
+			this.opacity = ( ( AdjustableOpacityColorConverter ) sourceAndConverter.getConverter() ).getOpacity();
+		}
+
+		this.lut = segmentationDisplay.tableViewer.getColoringLUTName();
```
tischi, line 102: @K-Meech I would prefer to fetch this information from `segmentationDisplay.coloringModel`? Can we get it from there? I think so... if not yet, we can add getter methods to MoBIEColoringModel.

tischi (commented):
Great! I had a few comments.

src/main/java/de/embl/cba/mobie2/display/SegmentationDisplay.java
K-Meech: I moved the lut getter from the tableViewer to the coloringModel - so this should all work now.

K-Meech:
Thanks for the comments @tischi! I made the changes you suggested

src/main/java/de/embl/cba/mobie2/display/SegmentationDisplay.java
```diff
@@ -75,7 +78,50 @@ public boolean showScatterPlot()
...
+
+		final SourceAndConverter< ? > sourceAndConverter = segmentationDisplay.sourceAndConverters.get( 0 );
+
+		if( sourceAndConverter.getConverter() instanceof AdjustableOpacityColorConverter)
+		{
+			this.opacity = ( ( AdjustableOpacityColorConverter ) sourceAndConverter.getConverter() ).getOpacity();
+		}
+
+		this.lut = segmentationDisplay.coloringModel.getColoringLUTName();
+		this.colorByColumn = segmentationDisplay.tableViewer.getColoringColumnName();
```
tischi, line 103: @K-Meech Also here we should use the coloringModel.

Something like:

```
if ( coloringModel.getColoringModel() instanceof CategoryTableRowColumnColoringModel )
{
    this.colorByColumn = coloringModel.getColoringModel().getColumnName();
}
```

src/main/java/de/embl/cba/mobie2/display/SegmentationDisplay.java
```diff
@@ -75,7 +78,50 @@ public boolean showScatterPlot()
...
+		}
+
+		final SourceAndConverter< ? > sourceAndConverter = segmentationDisplay.sourceAndConverters.get( 0 );
+
+		if( sourceAndConverter.getConverter() instanceof AdjustableOpacityColorConverter)
+		{
+			this.opacity = ( ( AdjustableOpacityColorConverter ) sourceAndConverter.getConverter() ).getOpacity();
+		}
+
+		this.lut = segmentationDisplay.coloringModel.getColoringLUTName();
```
tischi, line 102: @K-Meech 
There is for some of the coloring models already a method `getARGBLut()`
e.g. `CategoryTableRowColumnColoringModel` has this.
I think we should go into imagej-utils and add the method `getARGBLut()` to the other coloring models where we have it (e.g. `LazyCategoryColoringModel` `NumericTableRowColumnColoringModel`).
Then we do not need the `getColoringLUTName()`.
Do you want to do this?
I can also accept the PR and do it myself. No problem!

tischi (commented):
Thanks! Few more...

src/main/java/de/embl/cba/mobie2/display/SegmentationDisplay.java
K-Meech: Makes sense! If you had time to do this - that'd be great :)

PR #277, opened 2021-05-27
Project creator updates

K-Meech:
Main changes:
- Updates project creator for MoBIE2
- Allows overwriting existing views via 'Save current settings as view...' to e.g. change the default view
- Makes all Coloring luts use lower case to match the json schema ( internally before we were doing some conversions to upper case etc., but I think it's cleaner if we only support what the schema supports)

Needs a PR on imagej-utils too (I'll submit once this issue is resolved: https://github.com/tischi/imagej-utils/issues/48)


K-Meech:
@tischi Do you mind if for the moment we leave ```selectPathFromProjectOrFileSystem``` where it is? Then we can merge this + get everything back into the master branch. 
I'll have some time later in the week to do some refactoring and stop depending on ```selectPathFromProjectOrFileSystem```

tischi:
Sure, no problem!
Please merge if you think its ready otherwise.

K-Meech:
Ok - great! Can you merge the little PR I just made on imagej-utils? I don't think I have the permissions for that one...

K-Meech:
@tischi Should I move the remaining stuff from MoBIE to MoBIE2, so we can merge the whole playground branch back into master? Or do you want to?

tischi:
Please do it!

Then we should also delete the mobie package and rename the mobie2 into mobie :)

PR #281, opened 2021-05-31
merge packages

K-Meech:
Removes the 'mobie' package, and renames 'mobie2' back to 'mobie'

@tischi Can you take a quick look through, and check I haven't deleted anything you still need? I removed everything that wasn't used by any classes in MoBIE2.
I think there may have been some classes that are yet to be implemented in MoBIE2 e.g. the gene search for platybrowser that maybe you still want to keep for now? https://github.com/mobie/mobie-viewer-fiji/blob/playground/src/main/java/de/embl/cba/mobie/platybrowser/GeneSearch.java

Also, there are some old tests, that don't fit the new layout. For now, I just commented out the ones that had errors.


tischi:
I'll just merge. 
I made a branch playgroundPreMerge such that we can later find back the potentially missing classes easier.

PR #293, opened 2021-06-08
Improved support for different file formats

tischi:
I am working on supporting [this new spec](https://github.com/mobie/arabidopsis-root-lm-datasets/blob/6d4592803995abd0a4c029d1a75e2153220cc8f5/data/arabidopsis-root/dataset.json#L6).

There is a lot of code in the Project Creator that also needs to be changed. 
@K-Meech could you have a look at some point, further improving this branch?

### DIFF
--- a/src/main/java/de/embl/cba/mobie/source/ImageDataFormat.java
+++ b/src/main/java/de/embl/cba/mobie/source/ImageDataFormat.java
@@ -23,7 +23,7 @@ public enum ImageDataFormat
 			case OpenOrganelle:
 				return "openOrganelle";
 			default:
-				throw new UnsupportedOperationException( "Unknown image file format: " + this );
+				throw new UnsupportedOperationException( "Unknown file format: " + this );
 		}
 	}
 }
--- a/src/main/java/de/embl/cba/mobie/source/ImageDataFormat.java
+++ b/src/main/java/de/embl/cba/mobie/source/ImageDataFormat.java
@@ -1,0 +1,29 @@
+package de.embl.cba.mobie.source;
+
+import com.google.gson.annotations.SerializedName;
+
+public enum ImageDataFormat
+{
+	@SerializedName( "bdv.n5" )
+	BdvN5,
+	@SerializedName( "bdv.n5.s3" )
+	BdvN5S3,
+	@SerializedName( "openOrganelle" )
+	OpenOrganelle;
+
+	@Override
+	public String toString()
+	{
+		switch ( this )
+		{
+			case BdvN5:
+				return "bdv.n5";
+			case BdvN5S3:
+				return "bdv.n5.s3";
+			case OpenOrganelle:
+				return "openOrganelle";
+			default:
+				throw new UnsupportedOperationException( "Unknown image file format: " + this );
+		}
+	}
+}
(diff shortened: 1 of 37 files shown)


tischi:
@KateMoreva (@constantinpape) 
The changes in this PR may also affect your code (i.e. create merge conflicts).
But I think first @K-Meech should try to see how to make this work with the ProjectCreator and the Bookmarks.

K-Meech:
Sure - I have some time today, so I can take a look.
The main changes were the imageData/tableData in the sources right? Were there any changes to the xml files?

tischi:
imageData: changes (you will see in my PR branch)
tableData: changes (in fact I did not implement them yet, will do so now)
xml: no changes

tischi:
@constantinpape 
In the GridSourceTransforms the table data location would also change accordingly, right?

constantinpape:
> @constantinpape
> In the GridSourceTransforms the table data location would also change accordingly, right?

Yes.

tischi:
@K-Meech 
I also adapted the table data loading now (I could not test it because it does not compile it without your changes in the project creator).

K-Meech:
Ok - great. I'll look at the project creator now, and let you know when it's done.

K-Meech:
@tischi I fixed the issues with the project creator. So it should compile properly now.

src/main/java/de/embl/cba/mobie/source/ImageDataFormat.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.google.gson.annotations.SerializedName;
+
+public enum ImageDataFormat
+{
+	@SerializedName( "bdv.n5" )
+	BdvN5,
+	@SerializedName( "bdv.n5.s3" )
+	BdvN5S3,
+	@SerializedName( "openOrganelle" )
```
constantinpape, line 11: I think we should call this one "openOrganelle.s3":
That's in line with he bdv naming schemes and would allow to add a local version of the openOrganelle file format if we want to support this in the future.

constantinpape (commented):
Can we rename the serialized name from "openOrganelle" -> "openOrganelle.s3" and the internal name from "OpenOrganelle" -> "OpenOrganelleS3"?

src/main/java/de/embl/cba/mobie/source/ImageDataFormat.java
tischi: OK. done.

tischi:
Branch compiles and successfully opens the test project, like this:
```
new MoBIE("https://github.com/mobie/arabidopsis-root-lm-datasets",
 MoBIESettings.settings()
.gitProjectBranch( "new-data-spec3" )
.imageDataFormat( ImageDataFormat.BdvN5S3 ) );
```
Thus, it would be from my side ready to be merged into master. Should we? Since we anyway don't immediately ship the master to MoBIE-beta I think we could just do it.
Opinions?

constantinpape:
> Since we anyway don't immediately ship the master to MoBIE-beta I think we could just do it.
> Opinions?

Yes, let's do it. I can convert a few more of our projects later and put them on separate branches so that we have more test data.

PR #301, opened 2021-06-11
Loading columns

K-Meech:
Changes:
- Starts to implement loading extra columns on table
- renames remaining 'mobie2' references to 'mobie'
- stops view saving when columns are loaded from tables outside the project
- some minor project creator changes

@tischi Could you take a look? I'm running into an error with loading columns. e.g. if you open the remote zebrafish dataset, view any segmentation then load columns... from project, it errors with:
```
java.lang.IllegalArgumentException: Identifier not found
	at javax.swing.table.DefaultTableColumnModel.getColumnIndex(DefaultTableColumnModel.java:282)
	at de.embl.cba.tables.TableRows.setTableCell(TableRows.java:194)
	at de.embl.cba.mobie.table.TableViewer$1.cellChanged(TableViewer.java:156)
	at de.embl.cba.tables.tablerow.AbstractTableRow.lambda$notifyCellChangedListeners$0(AbstractTableRow.java:45)
	at java.lang.Thread.run(Thread.java:748)
```
Seems like there's some weird interaction with modifying the table after it has all its listeners instantiated...
Also, some refactoring will be needed to get this working for loading columns onto the grid table too.
(it might be quicker to just look through this together next week on zoom sometime?)

This also needs changes to imagej-utils (I'll put a draft PR on there in a minute)

### DIFF
--- a/src/main/java/de/embl/cba/mobie/MoBIE.java
+++ b/src/main/java/de/embl/cba/mobie/MoBIE.java
@@ -251,9 +251,21 @@ public class MoBIE
 		}
 	}
 
+	private String getRelativeTableLocation( SegmentationSource source ) {
+		return source.tableData.get( TableDataFormat.TabDelimitedFile ).relativePath;
+	}
+
+	public String getTablesDirectoryPath( SegmentationSource source ) {
+		return getTablesDirectoryPath( getRelativeTableLocation( source ) );
+	}
+
+	public String getTablesDirectoryPath( String relativeTableLocation ) {
+		return FileAndUrlUtils.combinePath( tableRoot, getDatasetName(), relativeTableLocation );
+	}
+
 	public String getDefaultTablePath( SegmentationSource source )
 	{
-		return getTablePath( source.tableData.get( TableDataFormat.TabDelimitedFile ).relativePath, "default.tsv" );
+		return getTablePath( getRelativeTableLocation( source ), "default.tsv" );
 	}
 
 	public String getDefaultTablePath( String relativeTableLocation )
--- a/src/main/java/de/embl/cba/mobie/MoBIE.java
+++ b/src/main/java/de/embl/cba/mobie/MoBIE.java
@@ -18,6 +18,7 @@ import de.embl.cba.tables.FileAndUrlUtils;
 import de.embl.cba.tables.TableColumns;
 import de.embl.cba.tables.TableRows;
 import de.embl.cba.tables.github.GitHubUtils;
+import de.embl.cba.tables.tablerow.TableRow;
 import de.embl.cba.tables.tablerow.TableRowImageSegment;
 import ij.IJ;
 import mpicbg.spim.data.SpimData;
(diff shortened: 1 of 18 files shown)


tischi:
> Seems like there's some weird interaction with modifying the table after it has all its listeners instantiated...

I think (hope) it is actually quite simple: the new columns do not exist yet in the JTable and thus they cannot be found.

I had a look in another repo where I am using the same code and there it looks like this:

```
private void registerAsTableRowsListener( TableRowsModel< T > tableModel )
	{
		for ( T tableRow : tableModel )
		{
			tableRow.listeners().add( new TableRowListener()
			{
				@Override
				public void cellChanged( String columnName, String value )
				{
					synchronized ( jTable )
					{
						if ( ! Tables.getColumnNames( jTable ).contains( columnName ) )
						{
							addColumnToJTable( columnName, value );
						}

						setTableCell( tableModel.indexOf( tableRow ), columnName, value, getjTable() );
					}
				}
			} );
		}
	}
```

Probably you could just copy that into MoBIE. 

Could you please try? 

Otherwise I can have a look next week.

K-Meech:
I'll give it a go! Could you send me the link to the repo? There are a couple of functions it relies on that aren't in mobie yet

tischi:
Cool! Thanks!

Here it is: https://github.com/tischi/segmentation-annotator/blob/classify/src/main/java/de/embl/cba/segmentationannotator/TableView.java

Beware that this is not the master branch, but a branch called "classify".
Also beware, that you probably should remove the other places where the JTable is updated, e.g. in terms of adding other columns, since now everything should go via the listening mechanism. 

I think that's better since I think our actual table model should be just the `List< TableRow >` and the JTable should be just considered a view on that table model, which automatically updates itself via the listening mechanism.

tischi:
More comments:
- There is obviously a terrible lot of  code duplication. The `TableViewer` exists in at least three different repositories almost identical. Once things stop moving one should think about refactoring to reduce code duplication.
- In the "classify" branch on the segmentation-annotator there is new functionality to train a random forest on annotated segments to predict the classes of the once that are not yet annotated (probably identical to the object classification implemented in ilastik). Obviously this should also be possible within MoBIE. Again, one has to think where this code should live. Maybe in imagej-utils....

tischi:
@K-Meech Let me know whether I should work on it. I think I would have some time this morning...

K-Meech:
@tischi That would be great! I didn't get a chance to look at it over the weekend, so nothing has changed since Friday

tischi:
@K-Meech 
Could you make this a branch within this repo? 
I think it may be complex for me pull and push to your fork...

K-Meech:
@tischi It's there as the branch 'load_columns' now https://github.com/mobie/mobie-viewer-fiji/tree/load_columns I'll do the same for imagej-utils

tischi:
OK, let me know once the ij-utils branch is there, then I'll pull both

K-Meech:
It's there now - 'load_columns' again: https://github.com/tischi/imagej-utils/tree/load_columns

tischi:
@K-Meech 
I changed some things and it works for me now.
I close this PR such that you can merge from the branch on MoBIE.

K-Meech:
Thanks @tischi. Would you have time to look at getting this working for grid tables too?

This issue is that the table merging code for grids is under ```GridOverlaySourceDisplay``` and not ```MoBIE``` as for the rest of the tables. (https://github.com/mobie/mobie-viewer-fiji/blob/load_columns/src/main/java/de/embl/cba/mobie/grid/GridOverlaySourceDisplay.java#L64) so I can't access it easily from 'load columns...' in the ```TableViewer```: https://github.com/mobie/mobie-viewer-fiji/blob/load_columns/src/main/java/de/embl/cba/mobie/table/TableViewer.java#L409

tischi:
Ok, I will work on it!

tischi:
I added this function to the MoBIE class:
```
public static void mergeImageTable( List< DefaultAnnotatedIntervalTableRow > intervalTableRows, Map< String, List< String > > columns )
```

Does that do the job for you?

You just need to load the additional table like this:

```
String tablePath = moBIE.getTablePath( tableDataFolder, relativeTablePath );
tablePath = Utils.resolveTablePath( tablePath );
Logger.log( "Opening table:\n" + tablePath );
final Map< String, List< String > > table = TableColumns.stringColumnsFromTableFile( tablePath ) );
```

K-Meech:
Looks good! I'll make the last few changes + merge it

PR #307, opened 2021-06-16
ViewerTransform improvements

tischi:


tischi:
@K-Meech this PR contains quite serious changes to the ViewerTransform changing stuff (fka BdvLocation).
Could you have a look and, if you have time, try to implement the serialization part of the new `ViewerTransformAdapter`, which you may need?

K-Meech:
Ok - sounds good! I'll have some time tomorrow to take a look through.

src/main/java/de/embl/cba/mobie/bdv/ViewerTransformLogger.java
```diff
@@ -0,0 +1,54 @@
...
+			// position
+			final PositionViewerTransform positionViewerTransform = new PositionViewerTransform( BdvUtils.getGlobalMouseCoordinates( bdvh ).positionAsDoubleArray() );
+
+			// affine
+			final AffineTransform3D affineTransform3D = new AffineTransform3D();
+			bdvh.getViewerPanel().state().getViewerTransform( affineTransform3D );
+			final AffineViewerTransform affineViewerTransform = new AffineViewerTransform( affineTransform3D.getRowPackedCopy() );
+
+			// normalized affine
+			final AffineTransform3D normalisedViewerTransform = Utils.createNormalisedViewerTransform( bdvh, Utils.getMousePosition( bdvh ) );
```
K-Meech, line 39: @tischi I wonder if we should use the central point of the viewer for this? Rather than ```Utils.getMousePosition()```. The issue is that we are doing this via a right click menu, so your mouse position depends on how far down in the list the command to log the view is.

K-Meech:
Also @tischi - I think the viewer transform classes need an additional timepoint parameter. See: https://github.com/mobie/mobie.github.io/blob/master/schema/view.schema.json#L485
There's an optional timepoint for ```affine```, ```normalizedAffine``` and ```position``` + also the option to only give the timepoint in the viewer transform (https://github.com/mobie/mobie.github.io/blob/master/schema/view.schema.json#L445)

tischi:
> I think the viewer transform classes need an additional timepoint parameter.

I could add it but on the JSON side the `timepoint` currently is specified outside the `viewerTransform`, is it? @constantinpape

src/main/java/de/embl/cba/mobie/bdv/ViewerTransformLogger.java
tischi: Good idea! I can change it...

constantinpape:
No it's specified as part of the viewer transform, see e.g. https://github.com/mobie/mobie.github.io/blob/master/schema/view.schema.json#L456-L468.
So a valid viewer transform would e.g. be
```
{
  "affine": [...],
  "timepoint": 1
}
```

K-Meech:
@constantinpape am I right that:
```
{
"timepoint":1
}
```
would also be a valid viewer transform? (due to https://github.com/mobie/mobie.github.io/blob/master/schema/view.schema.json#L445 )
If so, we would need another viewer transform class - something like ```TimepointViewerTransform```

constantinpape:
> @constantinpape am I right that:
> 
> ```
> {
> "timepoint":1
> }
> ```
> 
> would also be a valid viewer transform?

Yes, exactly.

K-Meech:
@constantinpape Could you maybe add a couple of extra bookmarks to the arabidopsis dataset?
- one with a normalised affine viewer transform + timepoint
- one with only a timepoint

Would be very useful for testing!

tischi:
Sounds great! Should I add the `timepoint` or will you do it @K-Meech ?

K-Meech:
@tischi - it'd be great if you could add the timepoint :) I can look into the ViewerTransform serialisation / checking it works with the project creator.

constantinpape:
I have added the example bookmarks now: https://github.com/mobie/arabidopsis-root-lm-datasets/blob/spec-v2/data/arabidopsis-root/dataset.json#L127-L141

Note that the normalized affine is more or less random, so I don't know if the resulting view will make sense. Feel free to replace the values with something more meaningful.

tischi:
> it'd be great if you could add the timepoint

done via https://github.com/mobie/mobie-viewer-fiji/pull/307/commits/65d727466f68c769b73ba0658c72db20b013bf83

K-Meech:
Ok @tischi - I added the serialisation (turned out to be a one liner!)
Also, I fixed a quick issue with the additional view loading.

tischi:
So we merge?

K-Meech:
Yes - sounds good! It's only the ```Utils.getMousePosition()``` change to go. I can make a separate issue for this.

tischi:
> It's only the Utils.getMousePosition() change to go. I can make a separate issue for this.

I will fix this and then merge.

Issue #329, opened 2021-06-30
Possible memory leak


K-Meech:
It seems that adding, and then removing views does not properly release memory

e.g. If I open the platynereis dataset, I start at a memory usage of about 38%. If I then view the bookmark 'Figure 2C: Muscle segmentation', it increases to 44%. Removing the cell segmentation after this, does not release the memory (remains at 44%)

This means viewing a lot of bookmarks in succession can result in an out of memory error.

tischi:
> can result in an out of memory error.

Does it really do result in an error? My experience is that Fiji sometimes hangs on to memory even though it is sort of released...

K-Meech:
I did get an out of memory error yesterday when I was checking the Platynereis bookmarks. I viewed pretty much all of them one after the other.
No solid numbers for this, but I got the impression that the more bookmarks that I viewed, the longer they took to load + the higher the memory usage.

tischi:
That's an interesting one!
If you want to increasing you developer capabilities, you could google for this: `visualvm memory leak Java` and try to find the leak using **visualvm**. It is fun! I already used it to find some leaks.

K-Meech:
Sounds good! I should have some time later in the week to take a look.

tischi:
Thanks!
See here: https://github.com/mobie/mobie-viewer-fiji/pull/334

K-Meech:
Ok - explicitly closing the image loaders definitely helps! I did the same thing (adding / removing the cell segmentation a few times) and now the threads are definitely being terminated. This also keeps the memory usage a lot lower than before, as it's clearing the caches too.
![Screenshot (304)](https://user-images.githubusercontent.com/24316371/124486972-0dc11780-dda6-11eb-9d3c-168d943771c2.png)

Memory usage does still seem to increase over time though... So it seems there are some other sources of leaks.

PR #334, opened 2021-07-05
Seal memory leaks

tischi:
@K-Meech 
Thank you for investigating the memory leaks.
Here is some code to get you started on testing whether closing the ImgLoader helps.


K-Meech:
Thanks! I'll take a look

K-Meech (approved):
@tischi I think we can already merge this now. It improves the performance a lot - see https://github.com/mobie/mobie-viewer-fiji/issues/329#issuecomment-874161932. There still seems to be some increase in memory over time, but I think we can tackle that later.

PR #336, opened 2021-07-06
Blending modes

tischi:
I am not sure yet whether everything is working in this branch because I still need to add blending modes to the yeast CLEM project dataset.json.

But it also contains the new `SelectedSegmentsColorConfiguratorCommand`. So, if you urgently need this, you could run from this branch (bad practice from my side by the way, I should have added above command to a new branch...).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
 
 		<!-- BIOP -->
 		<bigdataviewer-playground.version>0.3.4</bigdataviewer-playground.version>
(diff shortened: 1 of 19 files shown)


K-Meech (commented):
Thanks! The selected colour work great!

tischi:
@K-Meech 
I am merging this now.
Please be aware that there are now different dependencies.
Some most be build locally as they are still SNAPSHOT versions.

PR #342, opened 2021-07-07
Select by column values

K-Meech:
Allows selection of segments based on the values in a chosen column
- equal to some value ( for categorical and numeric columns )
- less than/greater than some value ( for numeric columns )


tischi:
Thanks a lot for working on this! Awesome!!

PR #347, opened 2021-07-09
Open ome zarr through MoBIE

KateMoreva:


constantinpape:
Awesome! @tischi I think it's best if you review this ;)

KateMoreva:
> Awesome! @tischi I think it's best if you review this ;)

Thank you, just a second, I messed up the last commit

tischi:
@KateMoreva 

Thanks  lot! Looks great!

1. What about adding an `AxesMapHelper` class with utility methods such as `boolean is4DC( Map< String, Int > axesMap`. Then I think we could get rid of the booleans in the `ImageLoader` class.
2. I am a bit confused about the file formats as there currently only seems to be `bdv.ome.zarr`. Does that mean we cannot yet read from S3? 

If you agree with the two issues above I would propose the following: you don't work on this anymore in this PR, but just make issues about this in the `mobie-fiji-viewer` repo and then we merge your PR as is, OK?  We can then work on those and other issues with new PRs against the bdv-ome-zarr repo, into which we want to move all this code, isn't it?

KateMoreva:
> @KateMoreva
> 
> Thanks lot! Looks great!
> 
> 1. What about adding an `AxesMapHelper` class with utility methods such as `boolean is4DC( Map< String, Int > axesMap`. Then I think we could get rid of the booleans in the `ImageLoader` class.
> 2. I am a bit confused about the file formats as there currently only seems to be `bdv.ome.zarr`. Does that mean we cannot yet read from S3?
> 
> If you agree with the two issues above I would propose the following: you don't work on this anymore in this PR, but just make issues about this in the `mobie-fiji-viewer` repo and then we merge your PR as is, OK? We can then work on those and other issues with new PRs against the bdv-ome-zarr repo, into which we want to move all this code, isn't it?

Thank you, ok.
Yes, the current format is ` bdv.ome.zarr`.
Where then I should work over implementing reading ome.zarr files (non bdv)?

tischi:
> Where then I should work over implementing reading ome.zarr files (non bdv)?

I would also do it for now in the bigdataviewer-ome-zarr repository.

constantinpape:
> Where then I should work over implementing reading ome.zarr files (non bdv)?

Yes, I also think that this is where it should go.

KateMoreva:
> > Where then I should work over implementing reading ome.zarr files (non bdv)?
> 
> Yes, I also think that this is where it should go.
`bigdataviewer-ome-zar`r then would have a dependency on MoBIE files. I thought the idea was not to have such one.

tischi:
> bigdataviewer-ome-zarr then would have a dependency on MoBIE files

Why? I agree that bdv-ome-zarr should not have mobie as a dependency; thus the next step is to move all the ome zarr stuff out of mobie-fiji-viewer into bdv-ome-zarr and have mobie-fiji-viewer depend on bdv-ome-zarr .

Or do you think there would still be an issue?

KateMoreva:
> > bigdataviewer-ome-zarr then would have a dependency on MoBIE files
> 
> Why? I agree that bdv-ome-zarr should not have mobie as a dependency; thus the next step is to move all the ome zarr stuff out of mobie-fiji-viewer into bdv-ome-zarr and have mobie-fiji-viewer depend on bdv-ome-zarr .
> 
> Or do you think there would still be an issue?

Yes, all the some.zarr files should be moved, and the repository contains all needed files (except new XmlN5OmeZarrImageLoader). The implementation of ome.zarr requires the changes in MoBIE and some other files, thus I think it should be made here.

PR #361, opened 2021-07-20
Ome.zarr.s3 opening with MoBIE

KateMoreva:


constantinpape (approved):
LGTM. The changes are small and I am pretty sure will work and not interfere with anything else, so please go ahead and merge if you have tested that it works.

Issue #352, opened 2021-07-15
When annotating, map number keys to categories


martinschorb:
It would be super handy to have the annotation categories mapped to the number keys (`1`- `0`) when the annotation window is active to massively speed up manual annotation.

tischi:
Good point!
I was 100% sure we have configurable keyboard shortcuts for the annotation classes....but indeed I cannot find it in the code anymore.
@K-Meech do you remember us having this functionality?

constantinpape:
For reference: https://github.com/tischi/imagej-utils/issues/28

K-Meech:
The corresponding commit for this is here: https://github.com/tischi/imagej-utils/commit/2ca807cc6cce512c3a133cb0972d38eb962c856b Not sure why this was removed in later versions...

tischi:
...at some point I decided to copy the Annotator code into MoBIE: https://github.com/mobie/mobie-viewer-fiji/blob/master/src/main/java/de/embl/cba/mobie/annotate/Annotator.java

I actually think once things stabilized more we could move it back to into imagej-utils (or some other repo).

So, I guess one would just have to copy the code from ij-utils to the mobie code. If you have time, please do it, otherwise I can do it (next week).

tischi:
> copy the code 

_Not everything_ as the code diverged quite a bit, but _only_ the part with the buttons and keyboard shortcuts.

tischi:
@K-Meech should I have a look into this? I may have some time this week 😄

K-Meech:
That would be great - thanks!!

tischi:
@vzinche Do you remember? I think we once had the option to annotate via keyboard shortcuts for each annotation category? Did you ever test this? For some reason we removed it again from the code and I do not remember why... maybe it did not work very well?

vzinche:
I don't remember discussing or testing this ever

PR #384, opened 2021-07-28
Keyboard shortcuts for the Annotator

tischi:
see: https://github.com/mobie/mobie-viewer-fiji/issues/352

@K-Meech 

I added back the code that was in the ij-utils commit. It turns out that this commit was just adding the button but did not implement the actual keyboard listener. I vaguely remember that I had trouble getting this to work, because the annotator typically is not the focused window, because people click in the BDV window. We could implement a "global" keyboard listener, but then if people write an e-mail in between it would still listen...

### DIFF
--- a/src/main/java/de/embl/cba/mobie/annotate/AnnotatedIntervalCreator.java
+++ b/src/main/java/de/embl/cba/mobie/annotate/AnnotatedIntervalCreator.java
@@ -13,30 +13,27 @@ import java.util.stream.Collectors;
 
 public class AnnotatedIntervalCreator
 {
-	private final Map< String, List< String > > columns;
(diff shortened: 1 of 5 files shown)


K-Meech:
Ah ok - makes sense. Maybe you could just have 2 listeners (one on the annotator + one on the bdv window), to avoid the global listener? Not sure how hard this would be!

tischi:
Yes, something like this. I guess that's why it was never really implemented because I found this a bit tedious 😉 , but I will give it a 2nd shot!

tischi:
I asked in the forum: https://forum.image.sc/t/java-keyboard-listener-for-multiple-windows/55771

PR #395, opened 2021-08-04
Implement shared queue for OMEZarr

tischi:
Hi @KateMoreva (cc @K-Meech),

I hacked in a `SharedQueue` for OMEZarr. This helps in terms of not creating too many threads when adding very many Sources to BDV.

I would like to merge this as it will enable @constantinpape to have a look at the plate view images.
But it would be good if you could take note of what I did here such that we can make a nice implementation of this in bdv-imageloader.

### DIFF
--- a/src/main/java/de/embl/cba/mobie/MoBIE.java
+++ b/src/main/java/de/embl/cba/mobie/MoBIE.java
@@ -1,5 +1,6 @@
 package de.embl.cba.mobie;
 
+import bdv.util.volatiles.SharedQueue;
 import bdv.viewer.SourceAndConverter;
 import de.embl.cba.bdv.utils.BdvUtils;
 import de.embl.cba.bdv.utils.Logger;
@@ -57,7 +58,8 @@ public class MoBIE
 {
 	public static final int N_THREADS = 8;
 	public static final String PROTOTYPE_DISPLAY_VALUE = "01234567890123456789";
-
+	public static final ExecutorService executorService = Executors.newFixedThreadPool( N_THREADS );
+	public static final SharedQueue sharedQueue = new SharedQueue(Math.max(1, Runtime.getRuntime().availableProcessors()));
 	private final String projectName;
 	private MoBIESettings settings;
 	private String datasetName;
@@ -69,6 +71,7 @@ public class MoBIE
 	private String imageRoot;
 	private String tableRoot;
 	private HashMap< String, ImgLoader > sourceNameToImgLoader;
+	private SourceAndConverterService sacService;
 	//private HashMap< String, SourceAndConverter< ? > > sourceNameToSourceAndConverter;
 
 	public MoBIE( String projectRoot ) throws IOException
@@ -85,6 +88,7 @@ public class MoBIE
 		PlaygroundPrefs.setSourceAndConverterUIVisibility( false );
 		project = new ProjectJsonParser().parseProject( FileAndUrlUtils.combinePath( projectRoot,  "project.json" ) );
 		this.settings = setImageDataFormat( projectLocation );
+		sacService = ( SourceAndConverterService ) SourceAndConverterServices.getSourceAndConverterService();
 		openDataset();
 	}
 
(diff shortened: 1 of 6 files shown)


src/main/java/de/embl/cba/mobie/MoBIE.java
```diff
@@ -57,7 +58,8 @@
 {
 	public static final int N_THREADS = 8;
 	public static final String PROTOTYPE_DISPLAY_VALUE = "01234567890123456789";
-
+	public static final ExecutorService executorService = Executors.newFixedThreadPool( N_THREADS );
+	public static final SharedQueue sharedQueue = new SharedQueue(Math.max(1, Runtime.getRuntime().availableProcessors()));
```
tischi, line 62: @KateMoreva this global queue will be used by all image loaders.

src/main/java/de/embl/cba/mobie/n5/zarr/N5OMEZarrImageLoader.java
```diff
@@ -119,6 +121,12 @@ public N5OMEZarrImageLoader(N5Reader n5Reader, HashMap<String, Integer> axesMap)
 		fetchSequenceDescriptionAndViewRegistrations();
 	}
 
+	public N5OMEZarrImageLoader(N5Reader n5Reader, HashMap<String, Integer> axesMap, BlockingFetchQueues< Callable< ? > > queue ) {
```
tischi, line 124: @KateMoreva here is a new constructor that can take this queue.

src/main/java/de/embl/cba/mobie/n5/zarr/N5OMEZarrImageLoader.java
```diff
@@ -291,10 +299,13 @@ private void open() {
 						maxNumLevels = Math.max(maxNumLevels, setupImgLoader.numMipmapLevels());
 					}
 
-					final int numFetcherThreads = Math.max(1, Runtime.getRuntime().availableProcessors());
-					final BlockingFetchQueues<Callable<?>> queue = new BlockingFetchQueues<>(maxNumLevels, numFetcherThreads);
-					fetchers = new FetcherThreads(queue, numFetcherThreads);
-					cache = new VolatileGlobalCellCache(queue);
+					if ( queue == null )
```
tischi, line 302: @KateMoreva  we only create a new queue if none is provided

src/main/java/de/embl/cba/mobie/n5/zarr/OMEZarrReader.java
```diff
@@ -38,7 +40,7 @@ private SpimData readFile() throws IOException
         N5OMEZarrImageLoader.logChunkLoading = logChunkLoading;
         N5OmeZarrReader reader = new N5OmeZarrReader(this.filePath, new GsonBuilder());
         HashMap<String, Integer> axesMap = reader.getAxes();
-        N5OMEZarrImageLoader imageLoader = new N5OMEZarrImageLoader(reader, axesMap);
+        N5OMEZarrImageLoader imageLoader = new N5OMEZarrImageLoader(reader, axesMap, MoBIE.sharedQueue);
```
tischi, line 43: @KateMoreva here the global shared queue is injected into the ImageLoader. We should enable this for all ImageLoaders, not only the OMEZarr ones.

tischi: Good point, I guess the functions in bdv-imageloader could have a way to accept a SharedQueue from whoever calls them. Or (less nice) there could be some global flag within bdv-imageloader to use a SharedQueue. I'd leave it up to you to come up with some good idea 😄  👍

KateMoreva: Ok. There will be no problems with accepting SharedQueue in loaders (they are called from MoBIE).

KateMoreva (approved):

Issue #397, opened 2021-08-05
Sources are added inactive/invisible in latest mobie beta


K-Meech:
On the latest mobie beta, new sources are added inactive/invisible. 
e.g. open the spec-v2 branch of the platybrowser, the raw data is added, but invisible. Manually toggling the 'active' checkbox in the bdv source settings shows it again.

constantinpape:
@tischi I assume the problem is that you set the `visible` default to false, but it should be true.

tischi:
not sure what's going on:

```
public abstract class SourceDisplay
{
	// Serialization
	protected String name;
	protected double opacity = 1.0;
	protected boolean visible = true;
```

@K-Meech maybe you could debug into it when you find some time?

K-Meech:
@tischi I'll take a look :)

PR #399, opened 2021-08-05
add default constructor

K-Meech:
Fixes https://github.com/mobie/mobie-viewer-fiji/issues/397

@tischi Ok! So it turns out, that if you don't have a default/empty constructor then gson.fromJson() sets any missing fields to their default values. i.e. null for objects, false for booleans etc.. It won't use the values you state as field=value.

I fixed this for the displays, but if there's anywhere else you rely on default values, you will need to add a constructor!


tischi:
wow! good find! not immediately intuitive...
Thanks!

PR #419, opened 2021-09-02
Update dependency

KateMoreva:


constantinpape (approved):
Looks good! One question: Is the corresponding version of `bigdataviewer-imageloader` released to scijava.maven already? If not, can we release it first and update the dependency in the pom.xml, so that MoBIE can be build with the version from scijava?

KateMoreva:
> Looks good! One question: Is the corresponding version of `bigdataviewer-imageloader` released to scijava.maven already? If not, can we release it first and update the dependency in the pom.xml, so that MoBIE can be build with the version from scijava?

Yes, it is released. MoBIE uses files from the latest version of  `bigdataviewer-imageloader` (after the reorganization of the repo )

constantinpape:
> Yes, it is released. MoBIE uses files from the latest version of `bigdataviewer-imageloader` (after the reorganization of the repo )

Awesome. Then I think this is def. good to go; @tischi can you have another look at the changes and merge if you think everything is ok?

KateMoreva:
> Looks good! One question: Is the corresponding version of `bigdataviewer-imageloader` released to scijava.maven already? If not, can we release it first and update the dependency in the pom.xml, so that MoBIE can be build with the version from scijava?

I also made a pre-release on Github.

K-Meech:
@tischi @KateMoreva Can this PR already be merged?

PR #425, opened 2021-09-08
Fix N5 and update bigdataviewer-vistools version

KateMoreva:


tischi:
@KateMoreva Why did you change the bdv-core dependency? We need this version and we also need that the name of the Source is changed during the transformation.

KateMoreva:
> @KateMoreva Why did you change the bdv-core dependency? We need this version and we also need that the name of the Source is changed during the transformation.

Ok, sorry

KateMoreva:
> > @KateMoreva Why did you change the bdv-core dependency? We need this version and we also need that the name of the Source is changed during the transformation.
> 
> Ok, sorry

Changes were reverted.

tischi:
Thanks!

PR #427, opened 2021-09-08
Add GitHub Actions CI

### LINKED ISSUE #303: Add basic CI
We have no CI at all here, which makes evaluating PRs a bit difficult.
I understand that this is not so easy for tests that involve the GUI (though I am sure there are some workarounds for this), but for starters having something that checks the build would already be very helpful.

imagejan:
Closes #303.


K-Meech:
Thanks @imagejan!! This looks great.
One note - I see you took out the hardcoded versions we had for various libraries e.g. bigdataviewer-core / imglib2 etc. I think we need these because the versions actually in imageJ (shipped with the Java-8 update site) aren't the same as those in the pom-scijava at the moment? @tischi is that right?

tischi:
>  think we need these because the versions actually in imageJ (shipped with the Java-8 update site) aren't the same as those in the pom-scijava at the moment?

This is correct, I think some of the versions are (unreleased) SNAPSHOT versions and we need them. This also means the current master cannot be build "in the cloud", because the dependencies we need are not yet on maven. @imagejan I guess this is an issue?

imagejan:
@K-Meech wrote:
> I see you took out the hardcoded versions we had for various libraries e.g. bigdataviewer-core / imglib2 etc.

Note that I only removed the (redundant) `<version>` tags, but left the versions in the `<properties>` section intact, so it should have no impact on the build.
But indeed, if the pinned versions are not available from the maven repository, the CI build will fail. But I think locally I was able to build it, so it should be fine 🤷

K-Meech:
Thanks @imagejan. But if it's listed in the <properties> and not in <version>, won't it still default to the version in the parent pom? I'm probably just not understanding maven correctly - but I didn't think it could automatically pick up versions from the properties?

imagejan:
@K-Meech the parent POM defines `<version>${some-artifact.version}</version>` for each dependency, so setting the property will override the version defined in the parent POM, that's the whole purpose of this way to declare the versions, and that's what makes @ctrueden's melting pot possible, to ensure compatibility of all Fiji components with each other.

K-Meech:
Ah ok - thanks!! I didn't know this - makes a lot more sense now. I'll merge now.

PR #432, opened 2021-09-13
fix group id

K-Meech:
I noticed that the current mobie-viewer-fiji pom doesn't list a 'groupId'. This means it is currently being deployed to the maven repository under a group of 'org.scijava'.

I changed it to match the groupId of ```bigdataviewer-imageloader```:  https://github.com/mobie/bigdataviewer-imageloader/blob/main/pom.xml#L10

@tischi Is this fine to merge?


tischi:
That's a good one. 

I am not sure....

Usually I do like this, e.g.:

```
<groupId>de.embl.cba</groupId>
<artifactId>bigDataProcessor2</artifactId>
<version>1.3.3</version>
```

so, maybe it could be 

```
<groupId>de.embl.cba.mobie</groupId>
<artifactId>mobie-viewer-fiji</artifactId>
<version>1.3.3</version>
```

or 

```
<groupId>de.embl.mobie</groupId>
<artifactId>mobie-viewer-fiji</artifactId>
<version>1.3.3</version>
```

@constantinpape @imagejan do you have an opinion?

tischi:
Hmm, but then again:

- https://github.com/bigdataviewer/bigdataviewer-playground/blob/824decf439653eda0a5d2d0acba3d55bbecd68a9/pom.xml#L12

@NicoKiaru what's your philosophy for the `group id`? I think you also sometimes use something with `ch.biop` don't you?

tischi:
yet another example:

- https://github.com/clij/clij2/blob/0bee5679c60618e3e4526da029eee24e2c2d3acf/pom.xml#L15

tischi:
...I am not sure `sc.fiji` would be ideal for us, because the mobie github organisation also contains python code unrelated to Fiji...

tischi:
Our code currently looks like:
image

According to this:
https://maven.apache.org/guides/mini/guide-naming-conventions.html

We should use 

```
<groupId>de.embl.cba.mobie</groupId>
<artifactId>mobie-viewer-fiji</artifactId>
<version>1.3.3</version>
```

I bdv-imageloader it looks like this...

image

Maybe we should refactor the code to add something between the `cba` and `n5`?

Then again, bdv, which has a `sc.fiji` group id does not follow this convention:

image

K-Meech:
I think 
```
<groupId>de.embl.cba.mobie</groupId>
<artifactId>mobie-viewer-fiji</artifactId>
<version>1.3.3</version>
 ```
would be a good solution.  Then bdv-imageloader could be refactored to have the same structure. Does it need the n5 on the end? Maybe it could also have folder structure src/main/java/de/embl/cba/mobie?

tischi:
### bdv-imageloader

> Then bdv-imageloader could be refactored to have the same structure. Does it need the n5 on the end? 

Maybe we could replace the `n5` with `mobie.io` and then change the group id to `de.embl.cba.mobie`. 

### mobie-fiji-viewer

In principle to make things even cleaner, we could then also add another package layer `viewer` into the code of mobie.

tischi:
### de vs org

Maybe to be more inclusive we should consider changing `de` to `org`.

image

### cba or not

I am also not sure we whether we could _remove_ `cba`, because I am not sure the name of the CBA might change in the future, whereas `org` and `embl` and `mobie` are very likely to stay for quite some time.

tischi:
Another argument: there have been many non CBA people involved.
Thus, my current favorite group id: `org.embl.mobie` and with package structures:
`org.embl.mobie.viewer`
`org.embl.mobie.io`

K-Meech:
Sounds good to me! That seems most inclusive / least likely to change in future.

KateMoreva:
Ok, I'll rename groupId to o`rg.embl.mobie `and package structure to `org.embl.mobie.io `in` bigdataviewer-imageloader `then

tischi:
Then we could discuss whether we want to rename the artifact id `bdv-imgloader` to `mobie-io`. I think this is nice and concise and would not clash with whatever Tobias might do in the future (e.g. `bigdataviewer-io`).

imagejan:
> do you have an opinion?

I agree `org.embl.mobie` is a good choice. The main part of the groupId should be the inverse of a domain that you control, or a domain of an institute you're working for, so embl.org is good.

`sc.fiji` should (ideally) only be used by plugins that _actually ship_ with Fiji by default, athough I don't think this is true for some plugins currently.

Also, it's a good idea (for findability of source code) if the mapping of Maven groupId to GitHub organization is a 1-to-1 mapping, so if you have control of a domain like `mobie.org`, a groupId of `org.mobie` would be more concise and tell me immediately that I can find the source in the `mobie` organization on GitHub 🙂

(I think similar discussions had been going on for the various parts of BDV, with its own GitHub org `bigdataviewer` and the bigdataviewer.org domain. But at some point it's too late to change artifact names and package structures, without breaking code using these artefacts in the wild...)

K-Meech:
Ok! GroupId is now ```org.embl.mobie``` with artifactId ```mobie-viewer-fiji```. Package structure is now ```org.embl.mobie.viewer```
Happy to merge @tischi ?

Issue #441, opened 2021-09-14
Project Creator: Display settings are not stored


tischi:
@K-Meech 

Is seems as if the display settings (min max) of the currently open image are not properly transferred to the MoBIE project:

<img width="748" alt="image" src="https://user-images.githubusercontent.com/2157566/133251322-f310c52d-d7c2-45ed-8774-9c3b96090c95.png">

K-Meech:
Yes - currently it just sets to the defaults for the bitdepth i.e. 8-bit = 0 and 255 as the limits.
I'll look into keeping the current settings.

tischi:
We should also fetch the color of the LUT.

K-Meech:
I'm not sure if fetching the LUT is worth it. We only support a small number (and most of these aren't in base fiji, I think?). So it would only be glasbey that could be matched?

tischi:
I don't mean multi-color LUTs at this point, I mainly mean just the color, e.g. `gray` or `blue` or the RGB values. I am pretty sure the `ImagePlus` has a getter method for something like this.

K-Meech:
Ah ok - I'll take a look!

Issue #440, opened 2021-09-14
Create Project with OME-ZARR: What happens to voxel size?


tischi:
@K-Meech When creating a MoBIE project and saving the image as ome-zarr (without the bdv) does that mean that the voxel sizes are lost? Because the current ome-zarr spec does not yet contain the transformations? If that's the case we should maybe disable this option until this is fixed?!

K-Meech:
Yes - at the moment the voxel size info is lost! (as there's nowhere to store it in the current spec)
I can disable it. Otherwise, we could add a warning that voxel size isn't retained for plain ome-zarr?

tischi:
> Otherwise, we could add a warning that voxel size isn't retained for plain ome-zarr?

I would disable it for now.

PR #445, opened 2021-09-14
Project creator changes

K-Meech:
Main changes:
- disables plain ome-zarr from all project creator menus (the functionality is still kept, so we can easily re-enable this once voxel sizes/transforms are fully supported)
- Keeps brightness/contrast settings for images
- Keeps single colour LUT for images >> e.g. 'Green'/'Magenta' etc... Note: at the moment this will give you weird results if your image is being viewed with a multi-colour LUT like e.g. 'Ice' - it will set the colour to be whatever the colour is for the max pixel value. @tischi - is there a better way to handle this? I couldn't find an easy way to figure out if a particular LUT is for a single colour or not?


K-Meech:
For https://github.com/mobie/mobie-viewer-fiji/issues/441 and https://github.com/mobie/mobie-viewer-fiji/issues/440

Issue #437, opened 2021-09-14
Command names capitalisation


tischi:
Currently this is not consistent:

<img width="307" alt="image" src="https://user-images.githubusercontent.com/2157566/133234751-9f265785-4956-46d9-be36-af709b379984.png">

I had a look and within Fiji it seems to be common to capitalize all words. 
So, for example, we would do "Set Blending Mode".
I suggest we change everything according to this convention.
@K-Meech OK?

K-Meech:
Yes - sounds good!

KateMoreva:
> Currently this is not consistent:
> 
> <img alt="image" width="307" src="https://user-images.githubusercontent.com/2157566/133234751-9f265785-4956-46d9-be36-af709b379984.png">
> 
> I had a look and within Fiji it seems to be common to capitalize all words.
> So, for example, we would do "Set Blending Mode".
> I suggest we change everything according to this convention.
> @K-Meech OK?

We also have here :
**Create or edit grid view
Show view description**

KateMoreva:
@tischi, @K-Meech should I change all of them, as my changes are in the same file?

tischi:
I found more: 

<img width="562" alt="image" src="https://user-images.githubusercontent.com/2157566/133237537-7d7c5de8-92e9-411e-a980-317dab97d384.png">

@KateMoreva Sure, go ahead and change whatever is within your reach 😃

KateMoreva:
Done in https://github.com/mobie/mobie-viewer-fiji/pull/452

PR #452, opened 2021-09-20
Change commands naming

KateMoreva:
- Change naming according to https://github.com/mobie/mobie-viewer-fiji/issues/437
- Clean main methods for OmeZarr commands


tischi:
Thanks a lot!!

Issue #454, opened 2021-09-20
beta: Save current Settings as view - cannot load such a view properly


martinschorb:
When loading back a view saved through the right click functionality, I have the following issues (all using the filesystem to store the JSONs):

- 'Load additional view' does not do anything.
- Pasting the view into `dataset.json` leads to the entry showing up in the MoBIE GUI, but when loading it from there, I get:

```
Exception in thread "Thread-18" java.lang.NullPointerException
	at org.embl.mobie.viewer.bdv.view.ImageSliceView.adaptConverter(ImageSliceView.java:89)
	at org.embl.mobie.viewer.bdv.view.ImageSliceView.show(ImageSliceView.java:59)
	at org.embl.mobie.viewer.bdv.view.ImageSliceView.<init>(ImageSliceView.java:48)
	at org.embl.mobie.viewer.view.ViewManager.showImageDisplay(ViewManager.java:305)
	at org.embl.mobie.viewer.view.ViewManager.showSourceDisplay(ViewManager.java:272)
	at org.embl.mobie.viewer.view.ViewManager.show(ViewManager.java:227)
	at org.embl.mobie.viewer.ui.UserInterfaceHelper.lambda$null$1(UserInterfaceHelper.java:454)
	at java.lang.Thread.run(Thread.java:748)

```

To recreate:
https://github.com/mobie/clem-example-project/tree/spec-v2

the view:
https://github.com/mobie/clem-example-project/blob/a3fccf72e3cee9df4e493bb7c997ebd6e15295f8/data/hela/dataset.json#L884

the exported view:
https://github.com/mobie/clem-example-project/blob/spec-v2/data/hela/clem53.json

tischi:
@K-Meech Is it ok, if you have a look at this first?

K-Meech:
I'll take a look @tischi. Thanks for reporting @martinschorb!

K-Meech:
Ok - I found the issue. This happens because there are sources where the source name isn't the same as the xml file name e.g. here https://github.com/mobie/clem-example-project/blob/a3fccf72e3cee9df4e493bb7c997ebd6e15295f8/data/hela/dataset.json#L56. At the moment, I grab source names here: https://github.com/mobie/mobie-viewer-fiji/blob/master/src/main/java/org/embl/mobie/viewer/display/ImageSourceDisplay.java#L72 by getting the name directly from the spimSource. This will be == the xml filename, and not the mobie source name. Hence the errors.

[...]

PR #457, opened 2021-09-22
View error

K-Meech:
For https://github.com/mobie/mobie-viewer-fiji/issues/454

- Changes ```List< SourceAndConverter< ? > > sourceAndConverters``` to ```Map< String, SourceAndConverter< ? > > sourceNameToSourceAndConverter``` in AbstractSourceDisplay. 
- Views now correctly load the mobie source name and not the name in the SpimData


tischi:
Thanks a lot!
Some things look now a bit more complicated than necessary, but I think it makes sense to use the same image data model ( `Map< Name , SAC >` ) throughout the project.

PR #458, opened 2021-09-23
Enable manual transformations

tischi:


K-Meech (approved):
Great - looks good! I'll work on the serialisation :)

PR #459, opened 2021-09-23
Serialise manual transforms

K-Meech:
Main changes:
- Added support for AnnotatedIntervalAdapter to view saving
- Save manual transforms into a new affine sourceTransformer called 'manualTransform'. This is done by wrapping each source (after any sourcetransformers have been applied) into another TransformedSource - this means the manual transforms can be retrieved separately from any other transforms. Note: this probably doesn't properly support crops / grids etc right now!


tischi:
I tried it and it works! Awesome! I have to admit I don't understand all of the logic of our code anymore, but as long as it works... 😸

Issue #465, opened 2021-09-24
Wrapping of Sources into TransformedSources


tischi:
@K-Meech 

I just now understood that you are now also wrapping here:
https://github.com/mobie/mobie-viewer-fiji/blob/9553a097624ed40df48124e9ab140c705c7f1189/src/main/java/org/embl/mobie/viewer/view/ViewManager.java#L250

Then I guess we can remove it here?!
https://github.com/mobie/mobie-viewer-fiji/blob/9553a097624ed40df48124e9ab140c705c7f1189/src/main/java/org/embl/mobie/viewer/MoBIE.java#L305

K-Meech:
Yes - I guess so! I wrap all the sources (after any sourceTransformers are applied), so it should still work fine.

K-Meech:
I'll remove it, and make a PR

PR #467, opened 2021-09-27
remove extra wrap

K-Meech:
Closes https://github.com/mobie/mobie-viewer-fiji/issues/465
@tischi I removed the extra wrapping as a transformed source. I checked that we could still manually move sources and save/load a view. All worked fine :)


tischi:
Thanks a lot! The place where do did the wrapping makes much more sense than were I did it :)

Issue #468, opened 2021-09-28
project creator issue


martinschorb:
tested on current release version and beta.

- create new empty MoBIE project on file system
- add dataset
- add source -> bdv format image -> bdv.n5, link image, 'create view' checked

leads to a completely frozen Fiji on Mac OS in 95% of the cases. No file open dialog is shown.
From the optical appearance of the buttons and UI elements, it looks like
https://github.com/mobie/mobie-viewer-fiji/issues/314

K-Meech:
Thanks for reporting @martinschorb! Could you give a screenshot of the ui appearance? (It'll make it easier for me to figure out where the issue is)

martinschorb:
![Screenshot 2021-09-28 at 16 13 04](https://user-images.githubusercontent.com/35071867/135104686-71a12f79-e95d-41d8-9865-4fd8c0e42cfe.png)

That's how it looks when freezing. Happened today on both Rachel's and my Mac.

martinschorb:
and that's right after the 
> add source -> bdv format image -> bdv.n5, link image, 'create view' checked
dialogue.

martinschorb:
The whole Java process freezes and needs to be force-killed.

K-Meech:
@tischi - could you possibly debug this one? It's really hard for me to fix these issues, as they work fine on Windows and only error on Mac. @martinschorb is probably correct, that it is again something to do with the look and feel. It hangs at this line: https://github.com/mobie/mobie-viewer-fiji/blob/develop/src/main/java/org/embl/mobie/viewer/projectcreator/ui/ProjectsCreatorPanel.java#L519, so probably the JFileChooser in ```selectOpenPathFromFileSystem``` is not playing well with the current laf.

tischi:
@K-Meech I would suggest simply **not to** use the MoBIE LAF for the ProjectCreator **at all**. Would that be acceptable or does it look too ugly?

```
public ProjectsCreatorPanel ( File projectLocation ) throws IOException {
        // I would  suggest to remove below line
        UserInterfaceHelper.setMoBIESwingLookAndFeel();
```

K-Meech:
I'll take a look! I think we still need to choose a laf to enforce though, otherwise e.g. if you are using the project creator, then opening mobie to check it, then going back to the project creator... it is hard to know which laf will be used.

e.g. the 'Create new mobie project...' dialog looks like this if you open it straight away:
[...]

PR #469, opened 2021-09-29
Project creator laf

K-Meech:
For https://github.com/mobie/mobie-viewer-fiji/issues/468

- stores the laf when mobie is opened, and resets to this
- sets no laf for the project creator, so the default is used

This means menus/dialogs before and after using mobie now look the same on Windows (and hopefully on all other os!)
Project creator on windows:
Screenshot (354)

@tischi - does this look alright on mac too?

### DIFF
--- a/src/main/java/org/embl/mobie/viewer/ui/MoBIELookAndFeelToggler.java
+++ b/src/main/java/org/embl/mobie/viewer/ui/MoBIELookAndFeelToggler.java
@@ -8,11 +8,12 @@ public class MoBIELookAndFeelToggler {
 
     private static LookAndFeel before;
 
-    public static void storeCurrentLaf() {
+    private static void storeCurrentLaf() {
         before = UIManager.getLookAndFeel();
     }
 
     public static void setMoBIELaf() {
+        storeCurrentLaf();
         FlatLightLaf.install();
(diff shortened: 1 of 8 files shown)


src/main/java/org/embl/mobie/viewer/ui/MoBIELookAndFeelToggler.java
```diff
@@ -0,0 +1,33 @@
...
+public class MoBIELookAndFeelToggler {
+
+    private static LookAndFeel before;
+
+    public static void storeCurrentLaf() {
+        before = UIManager.getLookAndFeel();
+    }
+
+    public static void setMoBIELaf() {
+        FlatLightLaf.install();
```
tischi, line 16: @K-Meech I was actually thinking that here (before line 16) you _always_  call the `storeCurrentLAF` before changing to the MoBIE one. Then one does not have to think about doing it explicitly. What do you think?

K-Meech: Yes - that makes sense! I'll change it.

K-Meech: done!

tischi:
Thanks! I merged it!

Issue #470, opened 2021-09-29
Load additional views: add views vs. replace existing


martinschorb:
Dear all,

when you load additional views into MoBIE that already exist, a duplicated one will be added to the dropdown.
This becomes confusing when the "same" view is in there multiple times.

I see two options:
- adding a suffix (index) if a view of the same name already exists in the `uiSelectionGroup` dropdown list currently shown
- replace the existing view with the new one

K-Meech:
Thanks @martinschorb - I think replacing the existing view with the new one would be easiest. Do you agree @tischi? I'll start looking into fixing this.

martinschorb:
Also I noticed that when saving a view, all sources/views that have ever apopeared in the current BDV window over its lifetime (no matter if currently displayed or not) are included. This generates a very messy file...

example:
https://github.com/mobie/clem-example-project/blob/reg_group/data/hela/a3_reg.json

everything on `em-overview`, every source referring to `a2...` and `tomo53` are completely irrelevant and were not shown.
You can tell that from the `"sourceDisplays"`.
But still all the transformations for these sources are included in the export.

Feel free to open another dedicated issue in case you want things ordered.

K-Meech:
I split this off into another issue here: https://github.com/mobie/mobie-viewer-fiji/issues/471

K-Meech:
This issue is fixed on the current develop branch. Newly loaded views should replace old ones.

PR #472, opened 2021-09-29
Replace views of the same name

K-Meech:
Closes https://github.com/mobie/mobie-viewer-fiji/issues/470

- checks if a view is already present in a particular group dropdown. This stops 'loading additional views..' with views of the same name leading to multiple entries in the dropdown. Newly loaded views replace old ones.


K-Meech:
@tischi - are you fine to merge this? It's a quick fix to stop duplicate view names in the dropdown when 'loading additional views...'

PR #507, opened 2021-11-03
Update dependencies

tischi:
Here another try, but let's only merge once the voxel size issue is fixed.


tischi:
Hmm, the build fails, because it cannot find the imagej-utils dependency on maven. Maybe the old way of putting it there does not work anymore...?
Another reason to get the new way working...

tischi:
@K-Meech I would like to merge this today, could you have a brief look at it?

K-Meech:
Looks good! I'll merge it.

Issue #323, opened 2021-06-24
Support changing the resolution of objects shown in the 3d


K-Meech:
In the old mobie, there was an option in the context menu to change the resolution of the 3D objects. (and also to read it from this view parameter - https://github.com/mobie/mobie.github.io/blob/master/schema/view.schema.json#L61)

This is not yet supported in the new mobie.

K-Meech:
@tischi Could you also look at this one if you had time? I'm trying to show a lot of segments in the 3d viewer, so it'd be great to be able to reduce the resolution here.

K-Meech:
Also, just to note, this isn't exactly the same as it was in the old MoBIE. We agreed here (https://github.com/mobie/mobie.github.io/issues/2#issuecomment-803969585) to make resolution3dView an array of resolutions [x_resolution, y_resolution, z_resolution], rather than the single double that it was previously.

tischi:
Should I then choose the resolution level where the voxel spacing is less than or equal to the requested in all dimensions?

K-Meech:
Yes - sounds good!

tischi:
I do it now slightly more conservative: I pick the level where the spacings are equal to or larger than the requested.

tischi:
Works now, one can change, the resolution.

New feature: It will only do this for the segmentation sources that are under the mouse in BDV when right clicking to get the context menu.

<img width="496" alt="image" src="https://user-images.githubusercontent.com/2157566/124709555-e41a0480-defb-11eb-8553-d40b71c502f8.png">

K-Meech:
Thanks @tischi! Just re-opening for some minor issues I found while using this:
- currently, updating the resolution via the context menu does not update the resolution of meshes already shown in the 3D viewer
- if you provide a resolution that is too low, then it errors because there are no voxels left for that segmentation. It would be nice to  provide some message asking people to try a higher resolution.

tischi:
> currently, updating the resolution via the context menu does not update the resolution of meshes already shown in the 3D viewer

I am not sure we always want this... maybe some people want to show certain meshes with higher resolution?
Maybe a checkbox: redraw all segments?

> if you provide a resolution that is too low, then it errors because there are no voxels left for that segmentation. It would be nice to provide some message asking people to try a higher resolution.

Do you have an example for this? I don't think I have ever seen this error.

K-Meech:
A checkbox sounds like a good plan! 

You can get this error by e.g. 
[...]

PR #538, opened 2021-11-16
Volume rendering resolution

tischi:
@K-Meech 

Does this PR [fix this](https://github.com/mobie/mobie-viewer-fiji/issues/323)?

Don't merge yet, I also want to make the image volume rendering configurable...


K-Meech:
Yes - that fixes it, thanks!

tischi:
@K-Meech Can you play a bit with this? I added a right-click menu for changing the image volume rendering resolution...

K-Meech:
Looks good! Although, it seems like you can't enter decimal values, only whole numbers? Also, you have the units 'um' on both the segments and image volume rendering menus. I guess this is actually whatever physical unit the project is using? E.g. I know some people use nanometer for their projects so it might be a bit confusing if it isn't really converting to microns.

tischi:
Thanks @K-Meech !

Better now?

K-Meech (approved):
Looks good! Thanks @tischi! Go ahead and merge it :)

PR #543, opened 2021-11-19
Fixes #541

### LINKED ISSUE #541: Navigation does not work for segmentation table in mergedGrid
Navigating to segments via selecting rows in the tables or segments in the image does not work for segmentations in a merged grid. There are two issues:
- when selecting a segment the viewer navigates to a wrong location (outside the image space), see second screenshot below
- the selection applies to all cells with the same id, not just the cell with this id in the image at hand. we somehow need to take the image-id into account here.

The example is from https://github.com/mobie/covid-if-project using the "cell segmentation".

aaa-nav1
aaa-nav2
aaa-nav3

tischi:
@K-Meech 

Could you have a look whether the code looks OK?

And, only if you have time, whether it really fixes https://github.com/mobie/mobie-viewer-fiji/issues/541?

For me, it does.


tischi:
wrong branch...

tischi:
Don't look at this one. This was wrong.

PR #544, opened 2021-11-19
Fixes #541

### LINKED ISSUE #541: Navigation does not work for segmentation table in mergedGrid

tischi:
@K-Meech

Could you have a look whether the code looks OK?

And, only if you have time, whether it really fixes #541?

For me, it does.

### DIFF
--- a/src/main/java/org/embl/mobie/viewer/MoBIE.java
+++ b/src/main/java/org/embl/mobie/viewer/MoBIE.java
@@ -202,6 +202,7 @@ public class MoBIE
 		sourceNameToSourceAndConverter = new ConcurrentHashMap<>();
 		setDatasetName( datasetName );
 		dataset = new DatasetJsonParser().parseDataset( getDatasetPath( "dataset.json" ) );
+		//dataset.sources
 
 		userInterface = new UserInterface( this );
 		viewManager = new ViewManager( this, userInterface, dataset.is2D, dataset.timepoints );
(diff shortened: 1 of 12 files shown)


K-Meech (commented):
@tischi Not sure if these are separate issues - but I see that ctrl + left click in the table doesn't highlight the corresponding segment in the viewer. Also, you can't ctrl + left click in the viewer to select a segment. 
Left clicking in the table doesn't go outside the bounds of the image anymore - but I'm not totally sure it's going to the right place? It doesn't seem like the colours in the table match the colour of the segment it focuses on?

tischi:
@K-Meech 

Thanks!

The cells should now have the correct colors.

It should now be possible to select them.

It still moves to the wrong place, but we will fix this later, it may be wrong anchor coordinates in the table.

Does it work for you? OK to merge? 

If it is OK, could you (i) please merge it and (ii) also refresh MoBIE-beta?
❤️

K-Meech:
Looks good to me! I'll merge + update mobie beta. I'll post in the issue when it's done :)

PR #547, opened 2021-11-22
Fixes #540

### LINKED ISSUE #540: MoBIE opens tables for segmentations that are not selected
When opening a view that has a segmentation with a table, the table is always opened, even if the segmentation is **not** shown by default in the view.
This behaviour is weird, but might be due to the fact that we do not have a "showTable" field in the spec for the segmentationDisplay.

@tischi: I think there are two ways to fix this:
- when opening a view with a segmentation show the tables only if the segmentation is selected (i.e. "visible:True")
- add an optional field "showTable" to the spec (default: True). Then it's the responsibility of the project creator to make sure that showTable is set to false if the segmentation is not shown by default. With this approach it would also be feasible to show a segmentation, but not the table when opening a view.

I would be fine with either solution, let me know what you prefer. (And in the second case I could update the spec and the example below.)

Example: 
- Open https://github.com/mobie/covid-if-project
- This opens the tables for "cell segmentation" and "nucleus segmentation" although these sources are not set to visible in the view:
aaaaaa-mobie-tables

tischi:
Now against the correct branch.

### DIFF
--- a/src/main/java/org/embl/mobie/viewer/display/AbstractSourceDisplay.java
+++ b/src/main/java/org/embl/mobie/viewer/display/AbstractSourceDisplay.java
@@ -22,10 +22,7 @@ public abstract class AbstractSourceDisplay implements SourceDisplay
 	{
 		return opacity;
 	}
-	public boolean isVisible()
-	{
-		return visible;
-	}
+	public boolean isVisible() { return visible; }
 
 	// Runtime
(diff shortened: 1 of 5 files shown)


K-Meech (approved):
Looks good to me @tischi!

tischi:
@K-Meech Thanks for checking it. Sorry, I just now had an idea how to make it even better. Can you have another look?

K-Meech (approved):
Yep - looks good! 👍

PR #564, opened 2022-01-09
Spim data opener ref

KateMoreva:


src/main/java/org/embl/mobie/viewer/source/SpimDataOpener.java
```diff
@@ -71,7 +71,7 @@ public SpimData openSpimData(String imagePath, ImageDataFormat imageDataFormat,
                 spimData = openOmeZarrData( imagePath, sharedQueue );
                 break;
             case OmeZarrS3:
-                spimData = openOmeZarrS3Data( imagePath );
+                spimData = openOmeZarrS3Data( imagePath, sharedQueue );
```
tischi: @KateMoreva But this would still be good to have!!
Can you somehow just add this functionality to the current develop branch in the SpimDataOpener in mobie-io?

KateMoreva: we have public `SpimData openSpimData( String imagePath, ImageDataFormat imageDataFormat ) {`
where  `spimData = openOmeZarrS3Data( imagePath );`

tischi: Yes, good point! Sorry!

Issue #585, opened 2022-01-25
Default image name


tischi:
@K-Meech 

Could you use `imp.getTitle()` to suggest an image name?

<img width="404" alt="image" src="https://user-images.githubusercontent.com/2157566/150987940-1d69d79e-2226-4df3-8677-0a5c2a68565b.png">

K-Meech:
Sure - good idea! I'll add this :)

Issue #571, opened 2022-01-18
Project Creator: exclusive views


tischi:
I think when creating a view for an image we need a checkbox whether the view should be exclusive or not. Right now it is not exclusive.

Issue #579, opened 2022-01-24
Project creator affine


tischi:
@K-Meech 

I guess the affine is **on top of** the voxel size calibration, is it? Could you make this more clear in the project create?

<img width="624" alt="image" src="https://user-images.githubusercontent.com/2157566/150812066-8e852e2d-28e6-4ba4-b9f2-3cd8ce57df8f.png">

K-Meech:
This was added quite early on - so it actually is directly the affine that goes in the xml i.e. including the voxel size calibration! It should probably be changed so it instead just adds it to the view

tischi:
Yes, I think so. Could change it like this?
And than maybe:` [ ] Add additional registration view`?
And below this we put the UI for entering the affine?

tischi:
Maybe better to leave it as it was but remove the option to create a view.
Let's just always create a view and if the affine is not the identity (you can check this on Java) add it to the view.

I would also suggest to have

```
row 1       1, 0, 0, 0
row 2       0, 1, 0, 0
row 3       0, 0, 1, 0
```

in the UI instead of the 12 numbers. Then it is very clear even without reading the help. 

What do you think?

K-Meech:
Changing the affine to three rows sounds good! Makes it easier to see what's going on

For the view, I'd prefer not to get rid of this option. Sometimes, if you want to do something complicated with e.g. grids. It might be useful to be able to add an image (and have it be converted / made into a source etc), but not have its own view. I.e. if for some reason you have images that you only ever want to be viewed together.

Maybe there could just be another checkbox like, ```[] Add affine transform to view``` with the 3 rows of numbers below it?

tischi:
> For the view, I'd prefer not to get rid of this option. Sometimes, if you want to do something complicated with e.g. grids. It might be useful to be able to add an image (and have it be converted / made into a source etc), but not have its own view. I.e. if for some reason you have images that you only ever want to be viewed together.

Yes, I thought about this as well, but you can't create a grid (or in fact any multi image) view with the project creator. I feel for the beginner users always having a view is good (they will otherwise wonder (i) "what is a view?" and (ii) "why can't I see my image?"). They could delete the view manually from the dataset.json if needed.

We could have a `Project Creator Expert Mode` at some point, but for the basic one I would keep it as simple as possible.

PR #588, opened 2022-01-27
Project creator fixes

K-Meech:
Main changes:
- suggests name from image title (fixes https://github.com/mobie/mobie-viewer-fiji/issues/585)
- adds option to make view exclusive (fixes https://github.com/mobie/mobie-viewer-fiji/issues/571)
- removes option to not make a view (for https://github.com/mobie/mobie-viewer-fiji/issues/579)
Screenshot 2022-01-27 101619


tischi:
Looks good!

Some more things:

- [ ] three rows for the affine
- [ ] one would need to be able to overwrite an image (e.g., now to fix the bugs in the `fillValue[]`). I think right now it just tells you it is not possible, does it? Better would be a "Are you sure?" window.

Add it here or we merge and make a new PR?

K-Meech:
Yes - happy to add these! But would be easier to do in another PR. I need to fix the issues with writing ome-zarr that came from the latest mobie-io changes first.
Could you make a separate issue for being able to overwrite images?

PR #593, opened 2022-02-03
Add back GeneSearch

tischi:


tischi:
@constantinpape someone from Arendt lab requested putting back the ProSPr gene search. However, we changed the names of the sources not to contain "prospr" anymore. I guess that means the only way for me to figure out what are the ProSPr sources is via the `uiSelectionGroup`, is it?

tischi:
Hold on... in the dataset.json the path still contains prospr...

tischi:
@K-Meech Could you please have a look? It is quite a hack, but before having more use-cases for project specific plugins I did not feel like developing a whole infrastructure for this....

constantinpape:
Just fyi, I think checking via the `uiSelectionGroup` would be a much cleaner solution than the filepath. (But I also agree that we don't need to put much effort into this before having a more general plugin system)

tischi:
> checking via the uiSelectionGroup would be a much cleaner solution 

You are right, I did this now: https://github.com/mobie/mobie-viewer-fiji/pull/593/commits/9b2885f0b5be5d680726f5fb6852da8e898a7f80

Another issue is that one should _not_ open "raw" prospr sources (as I currently do), but the views, which could contain additional transformations. Since the prospr sources are currently not transformed on the view level it is not yet critical and we could merge it as is and make a "bug" issue.

In fact, looking at the code, this is an interesting example of what we discussed the last days during the hackathon: In order to measure the local expression levels I currently iterate through some voxel space, which is not obvious when the source is transformed by more than a simple scale (I guess one would have to iterate through real space at some granularity?!). 

This is the current code, where I just take into account a _uniform calibration_ (aka voxel size aka scale) to determine the region in voxel space.

```
final HyperSphereShape sphereShape = new HyperSphereShape( ( int ) Math.ceil( radius / calibration ) );
final RandomAccessible< Neighborhood< T > > nra = sphereShape.neighborhoodsRandomAccessible( rai );
final RandomAccess< Neighborhood< T > > neighborhoodRandomAccess = nra.randomAccess();
neighborhoodRandomAccess.setPosition( getPixelPosition( position, calibration ) );
final Neighborhood< T > neighborhood = neighborhoodRandomAccess.get();
final Cursor< T > cursor = neighborhood.cursor();
while( cursor.hasNext() )
{
...
}
```

To make this work with more complex transformations looks quite a bit more involved and I would thus be for merging such that the Arendt's can use and making an issue.
@K-Meech @constantinpape OK?

src/main/java/org/embl/mobie/viewer/plugins/platybrowser/GeneSearchCommand.java
```diff
@@ -0,0 +1,56 @@
...
+	{
+		GeneSearchCommand.moBIE = moBIE;
+	}
+
+	@Override
+	public void run()
+	{
+		double[] micrometerPosition = new double[ 3 ];
+		final RealPoint realPoint = new RealPoint( 3 );
+		bdvHandle.getViewerPanel().getGlobalMouseCoordinates( realPoint );
```
K-Meech, line 42: Does this give you the right position? As you're selecting from the context menu, your mouse position will be different to where you originally clicked (especially since this is the last option in the menu)

K-Meech (commented):
@tischi - I agree, makes sense to expand it / make it more robust later.
I'm not sure if this is getting the right point at the moment though?

constantinpape:
> To make this work with more complex transformations looks quite a bit more involved and I would thus be for merging such that the Arendt's can use and making an issue.
> @K-Meech @constantinpape OK?

I agree, feel free to merge when you think the current version is working correctly.

src/main/java/org/embl/mobie/viewer/plugins/platybrowser/GeneSearchCommand.java
tischi: @K-Meech Amazing question! You really scared me there 😨 . However, it turns out, once the context menu pops up, the "focus" seems to be lost on the BDV window and thus the last position BDV remembers is the one where you clicked at 😅 .

To make sure and also for the user, I could add a BdvPointOverlay at the position where the user clicked for `GeneSearch`, what do you think? I think it would be nice.

K-Meech: Yes - sounds good! Would be nice to show exactly where the gene search is coming from :smile:

tischi: image

In fact, this is really fun now. With the new S3 storage and some multi-threading optimizations on my side this is really usable even from home.

Any more suggestions or should I merge and release to MoBIE-beta?

K-Meech: Looks cool! I'm seeing a different result on Windows though - for me, the circle appears at my mouse position after I select the 'search radius (micrometer)', not the original position. E.g. here I clicked at a point in the ventral nerve cord, and the circle appears much lower:
Screenshot (389)

PR #596, opened 2022-02-11
OmeZarr v4

KateMoreva:


KateMoreva:
@constantinpape here it is, all you need is to make sure that maven loaded the latest version of mobie-io (if `mvn clean` and `mvn install` fails, delete the library (it is usually in` /home/`USER_NAME`/.m2/repository/org/embl/mobie/mobie-io/1.2.1-SNAPSHOT` folder)

constantinpape:
Thanks @KateMoreva. This fixes the installation error.

Issue #586, opened 2022-01-25
multi-channel images


tischi:
@K-Meech 

What happens if I add a multi channel image?

K-Meech:
I'm not sure - I think it will try and convert the whole thing into an n5 etc file, including all channels. I guess for MoBIE, we need it to instead split the channels and add each as individual images?

tischi:
Yes, we have to split. 
Either we throw an error or we split automatically and auto-create image names with `_ch1`, `_ch2`, or something like this...
 
I actually tried it and I think it gets stuck (crash)...

K-Meech:
Ah ok - I'll try and auto-split the channels. I don't think it should be too difficult.

K-Meech:
@tischi I already have a working version of this that splits the multi-channel image, and adds each with the suffix _ch1, _ch2 etc. The issue is that this complicates the logic for being able to overwrite images.

E.g. say you write an image with 3 channels, as imageName_ch1, imageName_ch2, imageName_ch3. Then, you want to overwrite with the same imageName, but only 2 channels. Should it recognise the 3 existing channels by the name and delete them, before adding these 2? Or should it only replace the 2 channels, and leave the old third one alone?

Also, if you overwrite an image with channels with one that doesn't have channels. Should it recognise the existing channels based on the name and delete them? Or just go ahead and add this new image?
Thoughts?

tischi:
Good points!
Maybe then we just don't support this and show a more informative error?
`IJ.showMessage("You image ...; [ Image > Color > Split Channels]");`

K-Meech:
Yes, I think this is the easiest option for now. I'll make a PR to add this message.

PR #600, opened 2022-02-14
add message for multi-channel images

K-Meech:
fixes https://github.com/mobie/mobie-viewer-fiji/issues/586

I add a message in two places - so that it outputs whether you use it through the ui, or directly via the api.


tischi:
Thanks @K-Meech !

PR #601, opened 2022-02-14
Add project creator affine to view, not xml

K-Meech:
Fixes https://github.com/mobie/mobie-viewer-fiji/issues/590 and https://github.com/mobie/mobie-viewer-fiji/issues/579
Main changes:
- Splits affine transform in ui into 3 rows
- Adds this affine to the view, not directly to the xml. Now the xml will always contain the scaling from the image itself (from imp.getCalibration()), and this affine will be added to the view on top.
Screenshot 2022-02-14 160652

### DIFF
--- a/src/main/java/org/embl/mobie/viewer/projectcreator/DatasetJsonCreator.java
+++ b/src/main/java/org/embl/mobie/viewer/projectcreator/DatasetJsonCreator.java
@@ -1,5 +1,6 @@
 package org.embl.mobie.viewer.projectcreator;
 
+import net.imglib2.realtransform.AffineTransform3D;
 import org.embl.mobie.io.ImageDataFormat;
 import org.embl.mobie.viewer.TableColumnNames;
 import org.embl.mobie.viewer.Dataset;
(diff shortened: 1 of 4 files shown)


tischi:
@K-Meech Looking at this, I guess we do need to keep the BDV XML, otherwise we could not store the affine.
Unless we have somehow separate out the non-scale parts and put them into the view? 
Or how do we do this right now?

K-Meech:
@tischi - this currently places the scale (coming from the image settings under Image > Properties) into the xml, and once I've updated the v0.4 writers, into the ome-zarr metadata. The affine you specify in the project creator ui, is then **on top** of this, and is placed into the view. So you have to make sure the affine you give goes from physical space to physical space.

tischi:
Aha, so the affine that we show in the UI will **not** contain the voxel size?
This is different in the version that I am using right now. Right now it also contains the voxel size.

K-Meech:
Yes - exactly! It won't contain the voxel size. 
This is exactly what I change in this PR - the default affine used to include the voxel size, and write whatever you changed directly into the xml. Now, the affine is the identity by default, and is written to the view instead (the voxel size is then written to the xml).

K-Meech:
Maybe I need to add some more to the affine title to make this clearer? I'll explain it in the docs when I update them at least.

tischi:
OK, this is good, but then in the GUI, I think we should tell the users that we in fact captured the voxel size. 
Maybe add a `gd.addMessage("Voxel size: ...");`
And then change `Affine transform` to `Affine transform (on top of voxel size)` or `Additional view affine transform` or `Additional affine transform`  or something like this?

K-Meech:
Yes - sounds good!

K-Meech:
Ok! I added the voxel size message now, and changed the affine title to 'Additional view affine transform'
Screenshot 2022-02-15 111423

PR #603, opened 2022-02-16
V4 ome zarr

K-Meech:
Needs https://github.com/mobie/mobie-io/pull/69 first
- Removes bdv ome-zarr from project creator
- Adds v0.4 ome-zarr to project creator + ome-zarr export command


constantinpape:
https://github.com/mobie/mobie-io/pull/69 is merged now. @tischi could you give the changes here a try, so that we can move forward on the full release with ngff v0.4 support?

constantinpape:
Thanks for merging @tischi. Should we make a new beta release.

Issue #606, opened 2022-02-18
Save current view: add new view to UI


tischi:
@K-Meech 

Could you try to change the code such that `Save current view` also adds the view to the MoBIE UI? (1) I think this is more intuitive. (2) I would need this for this: https://github.com/mobie/mobie-viewer-fiji/issues/605, because otherwise one cannot access the cropped source.

K-Meech:
Yes - sounds good, I'll add this. Should this only happen when you save directly to the project, or also when you save to an arbitrary location on the filesystem?

tischi:
Thanks! 
I think it should always happen.

PR #607, opened 2022-02-19
Add MaskedSource

tischi:
Made great progress in enabling interactive cropping in MoBIE, even when the viewer is not aligned with the physical axes 🥳  

image

Also seems I found a much better way to implement the crop as a `MaskedSource` (thanks to great help in https://gitter.im/imglib/imglib2)

- [ ] implement the `center` feature, which is not yet done in the `MaskedSource` 
- [ ] check whether we could entirely replace the current `CropSourceTransformer` and make it it use the new `MaskedSource`, which seems to be much more performant and also easier to understand.
- [ ] implement https://github.com/mobie/mobie-viewer-fiji/issues/606

@martinschorb Could you please point me to a simple example where we use the crop of a source?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,7 +94,7 @@
 		<!-- EMBL CBA -->
 		<imagej-utils.version>0.6.4</imagej-utils.version>
 
-		<mobie-io.version>1.2.2</mobie-io.version>
+		<mobie-io.version>1.2.3</mobie-io.version>
 
 		<!-- Version 1.6.0-scijava-3-SNAPSHOT of j3dcore addresses a minor issue
 		https://github.com/fiji/3D_Viewer/issues/26
--- a/src/main/java/org/embl/mobie/viewer/MoBIEUtils.java
+++ b/src/main/java/org/embl/mobie/viewer/MoBIEUtils.java
@@ -368,7 +368,7 @@ public abstract class MoBIEUtils
 		view.preConcatenate( translate.inverse() );
 
 		// divide by window width
-		final int bdvWindowWidth = BdvUtils.getBdvWindowWidth( bdv );
+		final int bdvWindowWidth = bdv.getBdvHandle().getViewerPanel().getDisplay().getWidth();
 		final Scale3D scale = new Scale3D( 1.0 / bdvWindowWidth, 1.0 / bdvWindowWidth, 1.0 / bdvWindowWidth );
 		view.preConcatenate( scale );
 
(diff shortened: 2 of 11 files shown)


src/main/java/org/embl/mobie/viewer/command/CropSourcesCommand.java
```diff
@@ -64,6 +62,8 @@ public void run()
 				final SourceAndConverter cropSource = cropSource( maskInterval, maskTransform, sourceAndConverter );
 
 				// TODO: can we create a view for this?
+				final MoBIE moBIE = MoBIE.getInstance( bdvHandle );
```
tischi, line 65: @K-Meech Do you have experience here and could help? The point is to somehow create and register a new `view` for the `cropSource` (actually this is a `MaskedSource`, need to change names later, maybe).

tischi:
@K-Meech 
If you have some time it would be cool if you could have a look at this PR.
I [tagged you](https://github.com/mobie/mobie-viewer-fiji/pull/607/commits/002ce448c552086611f48ec4cbd4c1b3533d2c19) where I think you may be the expert 😃

tischi:
@K-Meech I think a method `viewFromSourceAndConverter` would be good to have, what do you think?

tischi:
I created a [PR](https://github.com/mobie/mobie-viewer-fiji/pull/608) for working on creating a view from a `sac`.

src/main/java/org/embl/mobie/viewer/command/CropSourcesCommand.java
K-Meech: I think you would need to create the relevant ```SourceTransformer``` by getting the min/max of the crop etc from the Source. Then somehow add it to the the ```currentSourceTransformers``` in the ```ViewManager```. Then, when you save the current settings as a view, it will be made into a view when this function is called: https://github.com/mobie/mobie-viewer-fiji/blob/develop/src/main/java/org/embl/mobie/viewer/view/ViewManager.java#L208

K-Meech:
For making a view form a sac, this is kindof what is being done by [here](https://github.com/mobie/mobie-viewer-fiji/blob/8216b509808319165cf2ac9cb1d64f89f588a465/src/main/java/org/embl/mobie/viewer/display/ImageSourceDisplay.java#L73) and [here](https://github.com/mobie/mobie-viewer-fiji/blob/8216b509808319165cf2ac9cb1d64f89f588a465/src/main/java/org/embl/mobie/viewer/display/AnnotatedRegionDisplay.java#L72). At the moment, this is triggered [here](https://github.com/mobie/mobie-viewer-fiji/blob/develop/src/main/java/org/embl/mobie/viewer/view/ViewManager.java#L219) and below, when new ```ImageSourceDisplay```, ```SegmentationSourceDisplay``` or ```AnnotatedIntervalDisplay``` are made. Some things are taken from the sac, and some from the display - so I'm not sure if you could get all the view settings directly from the sac?

tischi:
I am also not sure...

I think we'll simply have to give it a try in [this PR](https://github.com/mobie/mobie-viewer-fiji/pull/608)

Do you want to give it a go or shall I?

K-Meech:
Could you give it a go? I'm fixing up some project creator stuff at the moment.

tischi:
ok!

PR #618, opened 2022-02-23
Write views s3

tischi:
Needs https://github.com/mobie/mobie-io/pull/81

@K-Meech Could you please have a look at the changes I made to the `DatasetJsonParser`. Is that too naive? Looks somehow to easy to be true...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,7 +94,7 @@
 		<!-- EMBL CBA -->
 		<imagej-utils.version>0.6.4</imagej-utils.version>
 
-		<mobie-io.version>1.2.4</mobie-io.version>
+		<mobie-io.version>1.2.5</mobie-io.version>
 
 		<!-- Version 1.6.0-scijava-3-SNAPSHOT of j3dcore addresses a minor issue
 		https://github.com/fiji/3D_Viewer/issues/26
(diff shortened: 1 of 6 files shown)


KateMoreva:
@tischi, are you sure that the PR should be to the `main`, not to the `develop`? :)

tischi:
changed it to be against `develop`.

K-Meech (commented):
@tischi - I think that should work! Although, in ```saveDataset```, you would need to find another way to ```setIndent```. Gson uses tab by default, but we changed this to 2 spaces to match Constantin's python code. Maybe this can be set directly in the Gson though, rather than on the JsonWriter?

tischi:
@constantinpape @K-Meech 

> Gson uses tab by default, but we changed this to 2 spaces to match Constantin's python code.

looks tricky: https://github.com/google/gson/issues/1279

I think I would just do this: `gsonInstancePretty.toJson(object).replaceAll("  ", "\t");`

K-Meech:
@tischi - if the default for pretty printing is 2 spaces, then that's great! 2 spaces is what we want, not tabs :)

tischi:
Can we merge now?

constantinpape:
> Can we merge now?

No objection from the code side on my end; but could you check why the test is failing?

tischi:
Could not find the new mobie-io dependency yet.

PR #622, opened 2022-02-25
Fixes #617: Refactor to enable view saving to project via API

### LINKED ISSUE #617: ViewsSaver API
@K-Meech 

To automatically save views I would need three new public methods in `ViewsSaver`:

```
public boolean saveViewToProject( View view, String viewName, boolean overwrite ); // without UI, projection location should be automatically determined from moBIE.

public boolean saveViewToFileSystem( View view, String viewName, String path, boolean overwrite  ); // without UI

public boolean saveViewToFileSystem( View view, String viewName, boolean overwrite  ); // pops up UI to choose the file path
```

Could you please add those?

They should return `false` if `overwrite` was `false` and the `viewName` exists already and they could thus not save it.

Sounds good?

tischi:
Hi @K-Meech,

I think in fact for now I only need to be able to save the view to the project.
So this would suffice.

Looks OK for you?

### DIFF
--- a/src/main/java/org/embl/mobie/viewer/view/saving/ViewsSaver.java
+++ b/src/main/java/org/embl/mobie/viewer/view/saving/ViewsSaver.java
@@ -1,6 +1,5 @@
 package org.embl.mobie.viewer.view.saving;
 
-import org.embl.mobie.io.util.FileAndUrlUtils;
 import org.embl.mobie.viewer.MoBIESettings;
 import org.embl.mobie.viewer.Dataset;
 import org.embl.mobie.viewer.MoBIE;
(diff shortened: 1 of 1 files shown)


K-Meech (commented):
Looks good to me @tischi. It would be nice at some point to tidy up the naming a bit e.g. now there's a ```saveViewToDatasetJson``` (that can overwrite) and ```overwriteExsitingViewInDatasetJson``` (that also overwrites) which is a bit confusing. It's not a big deal though. 

Also, here you're throwing an exception when the view name exists, while everywhere else in the class it just logs a message and returns. It would be nice to be consistent with this. I'm not sure what's best here - I was avoiding errors popping up in the log for stuff like this (e.g. names already existing etc), as I get the impression that people don't read the red error messages (or think something is broken!), but they do read things that get printed. Anyway, again, it's not a big deal and could be fixed later.

tischi:
Yes, we should definitely clean this up.
But that's quite some work and we should then probably discuss and do this together.
Just let me know when you have some time for this.
Please make an issue.

PR #631, opened 2022-03-03
Remove all

tischi:


tischi:
@K-Meech 
Can we merge this?

K-Meech (approved):
Looks good to me @tischi! We'll just need to keep an eye on the memory usage, as now it's only closing the loaders when the clear button is pressed right?  This could lead to memory issues (like we saw before) if people don't use clear often...

tischi:
> We'll just need to keep an eye on the memory usage, as now it's only closing the loaders when the clear button is pressed right? 

yes, exactly.

constantinpape:
> > We'll just need to keep an eye on the memory usage, as now it's only closing the loaders when the clear button is pressed right?
> 
> yes, exactly.

Is this correct? Shouldn't it also close the other loaders when an exclusive source is added?

tischi:
> Is this correct? Shouldn't it also close the other loaders when an exclusive source is added?

In fact this is what the code does (below code is also used both when pressing clear or when showing an exclusive view:

```
	public synchronized void removeAllSourceDisplays()
	{
		// create a copy of the currently shown displays...
		final ArrayList< SourceDisplay > currentSourceDisplays = new ArrayList<>( this.currentSourceDisplays ) ;

		// ...such that we can remove the displays without
		// modifying the list that we iterate over
		for ( SourceDisplay sourceDisplay : currentSourceDisplays )
		{
			// removes display from all viewers and
			// also from the list of currently shown sourceDisplays
			// also close all ImgLoaders to free the cache
			removeSourceDisplay( sourceDisplay, true );
		}
	}
```

Issue #591, opened 2022-01-30
Create a project with Java scripting


tischi:
How are you doing in terms of Java API for the ProjectCreator?
It would be really good if one could create/edit a project by means of Java scripting (Groovy or Jython).

K-Meech:
Good point! Do I need to do anything extra to make it available via Groovy or Jython? I think most of the functions are already there on the Java side + I could easily add a page to mobie.github.io to show how to do this.

tischi:
You do not need to do anything special!

If the Java API works you just create an example script and test that it works in the Fiji script editor in MoBIE-beta.

This would be super useful for larger projects!

K-Meech:
I'm going to close this, as this is now possible on the latest MoBIE/MoBIE-beta - with details here: https://mobie.github.io/tutorials/scripting_project_creator.html

PR #632, opened 2022-03-03
project creator api javadoc

K-Meech:
For https://github.com/mobie/mobie-viewer-fiji/issues/591
Add javadoc for main project creator api functions

(I'll also update the mobie.github.io soon to give more examples)

### DIFF
--- a/src/main/java/org/embl/mobie/viewer/projectcreator/DatasetsCreator.java
+++ b/src/main/java/org/embl/mobie/viewer/projectcreator/DatasetsCreator.java
@@ -12,10 +12,18 @@ public class DatasetsCreator {
 
     private ProjectCreator projectCreator;
 
+    /**
+     * Make a datasetsCreator - includes all functions for creating and modifying datasets in projects
+     * @param projectCreator projectCreator
+     */
     public DatasetsCreator(ProjectCreator projectCreator ) {
         this.projectCreator = projectCreator;
     }
 
+    /**
+     * Create a new, empty dataset
+     * @param datasetName dataset name
+     */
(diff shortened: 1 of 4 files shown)


tischi:
Looks good @K-Meech !

One thought (@constantinpape): Do we actually support having images of mixed image formats in the same dataset? I have a feeling we specify on the project.json level the image format that we want to use throughout when loading images...

K-Meech:
I think that there can be multiple image formats in the same dataset (in terms of the spec), but this can't actually be opened in MoBIE at the moment, as it automatically chooses one format here: https://github.com/mobie/mobie-viewer-fiji/blob/develop/src/main/java/org/embl/mobie/viewer/MoBIE.java#L116 ?

tischi:
I would suggest to modify that project creator to just choose one image format in the beginning and then stick with it, what do you think? That would also make the API and UI easier.

Unless someone sees actual use-cases for mixed image data formats within a project?

@martinschorb @constantinpape ?

martinschorb:
> Unless someone sees actual use-cases for mixed image data formats within a project?

If you have two large volumes that have been part of a MoBIE project before and were converted to different formats - maybe because they have been collected at different institutes, one doing the X-Ray and one doing EM or Light Microscopy - and you then want to create one MoBIE project with both of them registered.

tischi:
Fair point @martinschorb.
OK, maybe then we leave it flexible (and, once necessary also implement opening such mixed datasets in MoBIE, I don't feel this should be a problem.......)

K-Meech:
Sounds good to me! I think leaving it flexible for now is fine. I'll go ahead and merge this.

Issue #623, opened 2022-02-25
Project 2D?


tischi:
@K-Meech 

I think we should give the user the choice whether a created project is 2D.

For example, I have a use-case where I start of with only 2D images, but then I would like to stack them in 3D (serial section TEM).

K-Meech:
I guess this could be chosen right at the start (when the project is first created). Then for a 3D project we allow 2D + 3D images to be added, and for 2D, only allow 2D.

K-Meech:
As the 'is2D' is specific for each dataset, I think I'll add this as a checkbox to the dialog for making a new dataset.

PR #635, opened 2022-03-04
Allow selection of 2D/3D when dataset is created

K-Meech:
fixes https://github.com/mobie/mobie-viewer-fiji/issues/623
Main changes:
- Dialog for creating dataset now has checkbox for 2D/3D
- When adding a 3D image to a 2D dataset, a dialog appears allowing you to change the dataset to 3D or abort
- adds tests for adding 2D images to 3D and 2D datasets
Screenshot 2022-03-04 124336

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
 
 	<groupId>org.embl.mobie</groupId>
 	<artifactId>mobie-viewer-fiji</artifactId>
-	<version>2.1.5</version>
+	<version>2.1.6</version>
 
 	<!-- mvn clean install -Dscijava.app.directory=/Users/tischer/Desktop/Fiji/Fiji-MoBIE-beta.app -->
 	<!-- mvn clean install -Dscijava.app.directory=/Users/tischer/Desktop/Fiji/Fiji-MoBIE.app -->
--- a/src/main/java/org/embl/mobie/viewer/projectcreator/DatasetJsonCreator.java
+++ b/src/main/java/org/embl/mobie/viewer/projectcreator/DatasetJsonCreator.java
@@ -32,11 +32,19 @@ public class DatasetJsonCreator {
         this.projectCreator = projectCreator;
     }
 
-    public void addImageToDatasetJson( String imageName, String datasetName,
-                                      String uiSelectionGroup, boolean is2D, int nTimepoints,
-                                      ImageDataFormat imageDataFormat, double[] contrastLimits, String colour,
-                                      boolean exclusive, AffineTransform3D sourceTransform ) {
-        Dataset dataset = fetchDataset( datasetName, is2D, nTimepoints );
+    public void addDataset( String datasetName, boolean is2D ) {
+        Dataset dataset = new Dataset();
+        dataset.sources = new HashMap<>();
+        dataset.views = new HashMap<>();
+        dataset.is2D = is2D;
+        writeDatasetJson( datasetName, dataset );
+    }
+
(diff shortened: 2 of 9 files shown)


tischi:
Maybe:

`[ ] Will all images be 2D and do you only want to render them 2D`?

Actually, because this is such a complex sentence, I wonder now, maybe we should handle this in the MoBIE viewer when opening a project?
For example, if the project contains the `is2D` flag, maybe we could ask: The project only contains 2D data, do you want to restrict rendering in BDV to 2D?
 I think that may be better.

@constantinpape @martinschorb  what do you think?

martinschorb:
BDVPG handles this upon opening a new BDV window and the decision 2D/3D will be linked to that window. No source you add to it (can be 3D sources to 2D window) will modify this.

I think in a similar fashion we could deal with it at the `view` level.

What happens when you add the first view when there is no BDV window open/active. It initiates a new BDV window. At this point the decision has to be made, correct?
So when running "remove all views" and then opening a 2D-only view, it would (after asking) open a 2D BDV window.

tischi:
> It initiates a new BDV window. At this point the decision has to be made, correct?

Yes, but we currently keep the same BDV window for one dataset. 
Thus, for us the decision has to be made on the dataset level.

K-Meech:
We could decide when each dataset is shown. It might be a bit annoying though if all your datasets only have 2D images, to keep on having to respond to this dialog. Also, there are only a few use-cases where people want to render projects with only 2D images in 3D - so maybe it's a bit confusing to ask every time?

tischi:
> Also, there are only a few use-cases where people want to render projects with only 2D images in 3D 

All serial section microscopy is like this, could be many projects?!

K-Meech:
True - but I mean there are also loads of examples where you don't want this (e.g. montages of one large TEM section). 

Also, it's something that you know when you create the dataset whether it should be displayed in 2D or 3D. E.g. if you have serial section microscopy images (with all the respective transformations in z in the views), then you never want to display this in 2D - right? So giving the choice every time MoBIE opens means a user can easily open a project they're not familiar with in a way that makes it hard to browse... (same for opening a project that should be in 2D - like a big 2D section - in 3D. As then you can rotate, and it's hard to tell what's going on...) What do you think?

constantinpape:
I agree with @K-Meech and would leave this at the dataset creation stage.

> `[ ] Will all images be 2D and do you only want to render them 2D`?

How about we just make it a bit easier: `Will all images be 2D and do you want to explore them only in 2D?` (wouldn't use "render" here)

tischi:
OK, then let's just add one of those long sentences to the dataset creator.
@K-Meech as a native  🏴󠁧󠁢󠁥󠁮󠁧󠁿  you may decide what to write there exactly 😄

K-Meech:
How about ```Limit images and display to only 2D?```. Or is this not explicit enough? (just trying to make it a bit shorter :smile: )

K-Meech:
I updated it to this in the PR:
Screenshot 2022-03-04 160335

tischi:
OK, great!

Do you then in the creator in fact also check that the images that are being added only 2D and throw an error if not?

K-Meech:
Yes! If you try and add a 3D image to a 2D dataset, it will pop up this dialog:
Screenshot 2022-03-04 153339
If you do it directly via the API (and not the user interface), it will just throw an error.

K-Meech:
I'll go ahead and merge this. If anyone still has things they want changed - then feel free to open an issue :smile:

Issue #639, opened 2022-03-08
Project Creator trouble


martinschorb:
Hi,

- Mac tested on 10.15.6 and 11.6
- new Fiji, MoBIE release
- Create/Edit MoBIE project
- add source
- bdv 
- link/move, click OK
- This should open a file browser to select the bdv xml
- instead all Fiji freezes

martinschorb:
Update:
all release versions we tried (Windows, Mac,...) refuse to load local projects created with the project creator.

K-Meech:
Thanks @martinschorb - do you see these issues with MoBIE-beta too? The main MoBIE release is quite far behind, so we might just need to do another release soon.

K-Meech:
Do you have the issue with opening local projects created with project creator too on MoBIE-beta? Or that's only on the release version?

For the file browsing, I can't reproduce this on Windows so it's a Mac specific issue. @tischi - would you have time to look into this with your mac? [This](https://github.com/mobie/mobie-viewer-fiji/blob/develop/src/main/java/org/embl/mobie/viewer/projectcreator/ui/ProjectsCreatorPanel.java#L614) line and [this](https://github.com/mobie/mobie-viewer-fiji/blob/develop/src/main/java/org/embl/mobie/viewer/projectcreator/ui/ProjectsCreatorPanel.java#L618) line are where the file browser should open for adding an n5 or ome-zarr image.

tischi:
I cannot reproduce this on my mac.

I created a new mobie project and then successfully opened it in three ways:
1. Within the Project Creator: View in MoBIE
2. Edit existing MoBIE project
3. Open MoBIE project

constantinpape:
@tischi: Did you use the latest MoBIE-beta version?

@martinschorb: Can you check if this issue still occurs with the latest MoBIE-beta? If yes, can you please make video to show how to exactly reproduce this?

tischi:
Ah, I see, it is about linking an existing image. I never did this. Also freezes for me. I will check.

tischi:
@K-Meech

If you google: JFileChooser hangs 
you find a lot.

Luckily I found some (non-obvious) fix in some other code of mine.
[Now it should work](https://github.com/mobie/mobie-viewer-fiji/commit/3f036f1e5724d43461a53f913f5d3072b6e8978a).
 
[...]

Issue #633, opened 2022-03-03
Move clear button down below `location` 


tischi:
I suggest moving the clear button down below `location` to just above the sources/view list. (in the center where there is the `...` now)

_Originally posted by @martinschorb in https://github.com/mobie/mobie-viewer-fiji/issues/630#issuecomment-1058127345_

K-Meech:
This is fixed on the current develop branch

PR #643, opened 2022-03-11
Refactor file chooser and move clear button

K-Meech:
fixes https://github.com/mobie/mobie-viewer-fiji/issues/639 and https://github.com/mobie/mobie-viewer-fiji/issues/633

- refactors use of JFileChooser to always use the workaround from @tischi (I had to add ```new Thread()``` in a few more places to stop the file choosers hanging on windows)
- changes to SNAPSHOT version
- moves clear button down below location (while I was working on the file choosers, I found a bug where loading/saving views leads to multiple copies of the clear button - so I thought I'd just go ahead and move it)
Screenshot 2022-03-11 115135

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
 
 	<groupId>org.embl.mobie</groupId>
 	<artifactId>mobie-viewer-fiji</artifactId>
-	<version>2.1.6</version>
+	<version>2.1.7-SNAPSHOT</version>
 
 	<!-- mvn clean install -Dscijava.app.directory=/Users/tischer/Desktop/Fiji/Fiji-MoBIE-beta.app -->
 	<!-- mvn clean install -Dscijava.app.directory=/Users/tischer/Desktop/Fiji/Fiji-MoBIE.app -->
--- a/src/main/java/org/embl/mobie/viewer/MoBIEUtils.java
+++ b/src/main/java/org/embl/mobie/viewer/MoBIEUtils.java
@@ -190,26 +190,38 @@ public abstract class MoBIEUtils
 
 	public static File lastSelectedDir;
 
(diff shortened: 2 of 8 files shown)


tischi:
@K-Meech 
Are you sure that the `new Thread(...)` is enough? I mean did you have a MacOS to test this? If it would have been that easy, I don't think I would have implemented all this complicated code...

I still think it would be better to have a dedicated class that let's us choose a file path or directory, otherwise we have to add `new Thread()` or something more complicated in many places. What do you think?

But if time is an issue we can also merge like this and hope for the best.

K-Meech:
Hi @tischi - I still do all the code you added! It's here: https://github.com/mobie/mobie-viewer-fiji/pull/643/files#diff-68b06c2441b8706520d9fe7c18020382052d1ec8d632f734c295dbdd16e66e7fR211 . So there's now a general ```selectFilePath``` and ```selectDirectoryPath```, that then call ```selectPath``` which is all of your code.

I had to add ```new Thread()``` on top of this to various places (e.g. the file dialogs for saving/loading views from the file system) otherwise it froze on Windows. I could try adding the new Thread() directly into your code though, so that it isn't in so many places?

K-Meech:
Ok - I had a quick go, and I think I can simplify it a bit. I'll move some things around.

K-Meech:
Ok, I couldn't find an easy fix, to get the ```new Thread()``` into the ```selectPath``` function without freezing up Windows. @tischi - are you happy to merge as is? I don't think I'll have time to find a better solution today.
It would be great if you could check that the project creator file chooser still works on mac + also if the file chooser in saving a view to the filesystem still works on mac.

tischi:
ok, cool.

PR #680, opened 2022-04-23
April2022

tischi:
Many changes. 
I can make a list if needed....


tischi:
@K-Meech 

I am still working on it...will ping you once it is finished.
It will be very hard to review though, because I changed so much...maybe we'll just have to merge it and hope for the best (given that all tests are passing)....sorry, but I had a limited time window and wanted to get as many things as possible done, and they depended on each other, so it felt too complex to do this with several PRs (probably possible, but not within my capabilities).

tischi:
@K-Meech 
I think I would like to merge this now into develop.
Reviewing the code changes is probably impossible ;-)
But maybe, if you have time, you could use this branch to open the `platybrowser` and click around a bit to see that I did not break anything major?

K-Meech:
@tischi - yeah, I think this is a bit big to review! I opened it quickly locally, and it seems to open fine. The window arrangement is a bit strange though - now the MoBIE window is all the way to the right. Is that intentional?
Screenshot (397)

tischi:
> now the MoBIE window is all the way to the right. Is that intentional?

Yes, sort of :) The issue was that one could not see the log window during the initial project loading, which is important. We can work on making this more beautiful again at some point.

Issue #698, opened 2022-05-03
What's here?


tischi:
In the CLEM example it is very hard to navigate and one would really like to know which source is where.

@martinschorb and I discussed adding a "What's here?" item to the context menu, which would print the sources under the mouse pointer to the log window (since one may need this quite often a keyboard shortcut would be good as well).

One alternative could be to constantly listen to the mouse pointer position and display the source names somewhere in the BDV window.

What do you think? @constantinpape @K-Meech?

constantinpape:
From a user perspective I would prefer not to create a new entry in the context menu and always show the name of the current source(s) under cursor somewhere in the bdv window.
But if that's harder too implement, or causes performance degradation, we can also go with the context menu entry.

tischi:
> But if that's harder too implement

I will have a look. I was mainly worried about cluttering the BDV window. 
I guess it would be nice to be able to turn this on and off.
Maybe one more context menu item for "Settings", into which we could combine some things?

K-Meech:
I'm not sure about displaying directly in the bdv window - if you have a lot of layers e.g. EM / LM / segmentations etc. it could be a lot of names to display. I kind of think the 'F' button + being able to toggle sources on and off is already enough? But I guess being able to display the sources at an exact position is useful too.

martinschorb:
> the name of the current source(s) under cursor somewhere in the bdv window.

This can be dozens, so it might be indeed cluttering wherever we show it..

tischi:
### Use case

<img width="1264" alt="image" src="https://user-images.githubusercontent.com/2157566/166416183-6b073e2a-ef9a-46a9-b226-a3ab25b779c3.png">

I want to change the brightness of the green fluorescent image currently shown in BDV. How do I find out which one it is? Currently one has to do trial and error (switching on and off) on the left, which is annoying (not practical), if there are many.

martinschorb:
> I want to change the brightness of the green fluorescent image currently shown in BDV.

What about changing the display in the sources list based on the viewer?
All sources that are currently not visible will be greyed-out (or smaller font,...) the ones that are shown stay normal and the ones that are underneath the current cursor position will turn bold.

martinschorb:
Maybe also the saturation of the background colour could change.

tischi:
I though of displaying an `*` next to the sources under the mouse cursor, but then we have this issue that when you move the mouse to change the color of one of those source, the star would disappear.

martinschorb:
[...]

Issue #681, opened 2022-04-23
Rename annotation_id column?


constantinpape:
we could also just always call it `entity`...

-----

<img width="248" alt="image" src="https://user-images.githubusercontent.com/2157566/164916076-701c9745-ab9e-4ec5-acf6-3653a28a85d8.png">

This is now a bit confusing (`annotation_id` and `Annotate`), because they have nothing to do with each other, but I think this will be hard to change now...

_Originally posted by @tischi in https://github.com/mobie/mobie-viewer-fiji/issues/666#issuecomment-1107534815_

constantinpape:
I agree that this is a bit confusing. I think on the python / spec side it would not be so much effort to change this and we only have a few projects that use the `sourceAnnotationDisplay` and associated tables, so updating these would also not be too much work.

tischi:
It would also be easy to change on the java side.

Ideas:

- `region_id`
- `source_id` (this would match better the `sourceAnnotationDisplay`)

I realize one thing that confuses me with `source_id` as well as `sourceAnnotationDisplay` is that one can annotate multiple sources (=images) with one table row. 
I think that's also something that @martinschorb stumbled over when trying to use the spec. It would be nice to find a terminology that makes it more clear that one can annotated multiple images...

`sources_id` is ugly....maybe `region_id`.

constantinpape:
I think `region_id` is good!

constantinpape:
done

PR #703, opened 2022-05-05
Diverse refactoring

tischi:
Here's some refactoring that I did / want to do.

@K-Meech @constantinpape 
You could have a brief look through the points if you want.
I don't think there is anything too critical though.

---
SourceDisplay interface:

- [ ] Add getMask()?
- [ ] Maybe getRealInterval would be sufficient?
  - This would allow checking whether a SourceDisplay is currently visible in BDV in terms of intersecting with the current view (without checking all the individual sources, which may be expensive)
  - For this to work correctly, the SourceDisplay would need to listen whether the Transformation of one of the containing sources was changed and then accordingly update the mask

--- 

- [X] Create a SliceView interface and AbstractSliceView
- [ ] SliceView and AbstractSliceView could contain the getRealInterval() method (s.a.) to indicate the union location of all sources that are displayed in this SliceView. Then SourceDisplay.getSliceView().getRealInterval() would be used.
- [X] Add close() method; maybe just one implementation in AbstractSliceView needed?
- [X] AbstractSliceView constructor should not take bdvHandle, but the SliceViewer, which could implement show(sac), which can be used to show sources and do common things.

---

Change names:

- [x] AnnotatedRegionDisplay -> AnnotationDisplay ?
- [x] AnnotatedSourceDisplay -> RegionDisplay ?
- [x] SegmentationSourceDisplay -> SegmentationDisplay ?
- [x] ImageSourceDisplay -> ImageDisplay ?

---

- [x] AnnotatedRegionDisplay.getSources() currently returns all annotated sources. Is this correct? In fact, AnnotatedRegionDisplay only displays one Source, namely the `SourceAndConverter< IntType > sourceAndConverter` which implements the overlay. I thus think it is wrong; also could not see an issue in the code that uses his method.

--- 

- [x] MoBIEHelper.containsAtLeastOne could be replaced by some Java 8 code.

---

In ViewManager, there is the code following:

`// remove any sourceTransformers, where none of its relevant 'sources' are displayed
...
`

- [x] Refactor into a method called updateCurrentSourceTransformers?
@K-Meech, OK?

--- 

- [x] BdvGlobalMousePositionProvider --> GlobalMousePositionProvider
- [x] BdvCircleOverlay --> CircleOverlay

---

- [X] PublishedProjectsCreator --> PublishedProjects

--- 

PublishedProjectsCreator:

- [X] Create method for each project for better readability.

---

- [x] *Helpers --> *Helper 
- [x] *Utils --> *Helper

In several places (to be consistent, also with bdv-pl)

---

- [X] ThreadUtils --> MultiThreading
- [X] ThreadUtils: n => num 

---

- [X] `AdditionalViews` and `AdditionalViewsLoader`: Move up one level up in package structure and delete `additionalviews` package.

---

- [x] OMEZarrViewer: Move up two levels up to org.embl.mobie

---

- [x] view.saving.* ---> view.save.*

---

- [x] ViewsSaver --> ViewSaver, @K-Meech, OK?


K-Meech:
Sounds good to me @tischi ! For renaming the displays, maybe it would be best to match the names in the json schema as much as possible? (https://github.com/mobie/mobie.github.io/blob/master/schema/view.schema.json)

constantinpape:
I just had a brief look at the changes and they overall look good to me.

> For renaming the displays, maybe it would be best to match the names in the json schema as much as possible? 

I agree that if we rename here we should also consider renaming in the jsoschema. And I also think that the new names are better. But renaming in the schema means that we also need to update all projects and that would be a bit of work for me and I would only have time for it maybe middle of next week. (and when we do it we should include the change from #681).

constantinpape:
@tischi Do you still want to go ahead with the renaming? I could probably implement this and change it for our projects on Wednesday or Thursday.

tischi:
> Do you still want to go ahead with the renaming?

Yes, do the names that I proposed sound good?

constantinpape:
> Yes, do the names that I proposed sound good?

Yep, sounds good. I will go ahead and implement this and ping you once there's a branch to try it on.

tischi:
I am done with the Refactoring.
I moved the open points to https://github.com/mobie/mobie-viewer-fiji/issues/698 and will work on it in https://github.com/mobie/mobie-viewer-fiji/pull/699 once this PR is merged.

@K-Meech 
Again, too many code changes to review this...sorry...I should get better at this...
All tests pass, so it should be fine.
Maybe play around with it a bit.
Let me know if we can merge.

Issue #721, opened 2022-05-15
Simplify segments rendering UI


tischi:
Should we merge those into one in the GUI?
- Change Opacity Of Non-selected Segments
- Configure Coloring of Selected Segments
- Configure Label Rendering

---

> Yes, I think that would be good. I think "Configure Label Rendering" is a good name for the combined GUI.

tischi:
Working on this here: https://github.com/mobie/mobie-viewer-fiji/pull/722

tischi:
@constantinpape 
We also have:  “Configure Segments Volume Rendering”.

1. Should we also merge this into the "Configure Label Rendering"?
2. Segments or Labels?

constantinpape:
> We also have: “Configure Segments Volume Rendering”.
> 
>     1. Should we also merge this into the "Configure Label Rendering"?

No, I wouldn't merge it; since this is responsible for a different window (i.e. for the 3d viewer instead of the bdv window).

But I agree that we should use the same name everywhere, my vote would be for "Label", i.e.
- "Configure Label Rendering"
- "Configure Label Volume Rendering"

PR #722, opened 2022-05-15
Issue721

tischi:


tischi:
@K-Meech 
This should fix https://github.com/mobie/mobie-viewer-fiji/issues/721.
Could you please check?

K-Meech:
@tischi - I don't think I'll have time to look today (I'm on the microscope). I can probably look tomorrow - if that's not too late? (same for your other PR)

K-Meech (approved):
Looks good to me @tischi ! We need to update the mobie.github.io to match these changes - I can do this. I'll make an issue on that repo.

PR #726, opened 2022-05-17
some new JSON specs

tischi:


tischi:
> I think the name change for "selectedAnnotationIds" to "selectedRegionIds" is missing.

done

constantinpape:
Do you want to go ahead and merge despite failing tests (I guess because most projects in the tests are not updated), or do you want to wait till I have test branches for everything and we make sure that it works for them?
(I could do all the updates later today)

tischi:
I think there anyway is so much changing right now that I will just merge it...

PR #727, opened 2022-05-17
Make randomColorSeed serialisable

tischi:
@K-Meech 

I implemented: `( ( CategoryColoringModel< ? > ) coloringModel ).getRandomSeed();`

I tried using this where we serialize the current view but I could not find it...

This class seems not to be used anywhere?! `ViewFromSourceAndConverterCreator`??


tischi:
I found it, it should go into: `AnnotationDisplay`.

But should we then remove: `ViewFromSourceAndConverterCreator` ?

K-Meech:
@tischi If it's not used anywhere, then go ahead and remove it. I think it was something you added anyway?

tischi:
> If it's not used anywhere, then go ahead and remove it. I think it was something you added anyway?

not sure now; i'll leave it in, maybe there was some idea behind...

Otherwise could you have a look whether this PR looks good to you?

K-Meech:
Can do! I'm making some videos for the paper at the moment, but I can look later

tischi:
> Can do! I'm making some videos for the paper at the moment, but I can look later

Thanks! You will probably just have to look at the code, because I merged breaking changes of the JSON spec into this PR. Not sure, maybe the platybrowser will still open with this.

K-Meech:
@tischi - on a very quick look, it seems good to me! But at the moment the version of imagej-utils you're using isn't available remotely (so the tests will fail). Could you release a new version of imagej-utils and use that instead?

tischi:
> Could you release a new version of imagej-utils and use that instead?

I will do this later.

PR #730, opened 2022-05-18
issue710: default location

tischi:


martinschorb:
We probably need a parser for this free text field (even without https://github.com/mobie/mobie-viewer-fiji/issues/731). It is very likely that users do not paste true and valid JSON strings.

tischi:
For now, I could just include a try-catch and throw an error if the JSON parsing fails.
We can then later add the parsing into the catch, OK?

PR #783, opened 2022-06-06
Issue782: refactor loading of additional tables

tischi:


tischi:
In principle this is done.
One could certainly do more to clean it up even more.
One should put this to MoBIE-beta for testing.

PR #789, opened 2022-06-08
issue786: LazySourceAndConverter

tischi:
`adaptPosition( position, selection.imageId() );` in `SegmentationSliceView` seems to be the only reason why we currently need to transform all the individual sources in the `MergedGridView`. If we could get rid of this, loading the sources would be much faster as this could be done lazily.


tischi:
Did a lot of it. 

What's left (apart from debugging) is to implement the actual lazy loading inside the lazySAC.

tischi:
There now is a lot of special code for `getName()`.

- [ ] Maybe a `lazySpimSource` within the `lazySourceAndConverter` would have been the better option...

Issue #305, opened 2021-06-14
Replacing List< TableRows > with a TableModel


K-Meech:
It would be good to replace ```List< TableRows > ``` , with a more complex TableModel class. i.e. one that could retain more information about where to load and store parts of itself.

This would simplify some parts of the table code e.g. loading columns from segmentation tables vs grid tables

tischi:
And also have a thread safe (`synchronized`) method to add table rows (trigger by lazy loading).

There are:

- `de.embl.cba.tables.tablerow.TableRowsModel`;
- `org.embl.mobie.viewer.table.TableRowsTableModel`;

We should look to merge the functionality.

PR #793, opened 2022-06-10
Issue400: Lazy table loading

tischi:


tischi:
- [ ] https://github.com/mobie/mobie-viewer-fiji/issues/305

tischi:
Next steps:
When opening a SegmentationSource consider to immediately wrap it to return its table rows as pixel values (then we can stitch it in the FunctionGridSource).
Question: what to do with the Converter? This needs a `coloringModel` that is only available later when the `SegmentationDisplay` is instantiated. 

In fact, for **both** `SegmenationDisplay` and `ImageDisplay` we are completely changing the `Converters` that are opened. 

Thus, probably for `mobie.openSourceAndConverter` could be replaced by `mobie.openSource()`? Ah, not really, because this a convenient way to keeping together the `Source` and `VolatileSource`... 

But, during opening we could replace the respective `Converter`s by some `PlaceHolderConverter`, which does nothing.

tischi:
Consider: configure `SourceAndConverter` as much as possible already during initial loading, for both `SpimSource` and the `Converter`, based on whether this is a `SegmentationSource` or `ImageSource`.

tischi:
e.g. put this into the loading:

```
	// TODO: Can't we do this already during loading?!
	private SourceAndConverter< ? > wrapConverterAsAdjustableOpacityConverter( SourceAndConverter< ? > sourceAndConverter )
	{
		final Converter< RealType, ARGBType > converter = ( Converter< RealType, ARGBType > ) sourceAndConverter.getConverter();
		final AdjustableOpacityColorConverter adjustableOpacityColorConverter = new AdjustableOpacityColorConverter( converter );
		final Converter< ? extends Volatile< ? >, ARGBType > volatileConverter = sourceAndConverter.asVolatile().getConverter();
		final VolatileAdjustableOpacityColorConverter volatileAdjustableOpacityColorConverter = new VolatileAdjustableOpacityColorConverter( volatileConverter );
		sourceAndConverter = new ConverterChanger( sourceAndConverter, adjustableOpacityColorConverter, volatileAdjustableOpacityColorConverter ).get();
		
		return sourceAndConverter;
	}
```

PR #806, opened 2022-06-28
Fix opening of file system projects without bdv.n5

### LINKED ISSUE #805: Only projects with bdv.n5 data can be loaded locally
otherwise the else in https://github.com/mobie/mobie-viewer-fiji/blob/72e7d6f6cab8aea1c222aff4e115b4b200e515f3/src/main/java/org/embl/mobie/viewer/MoBIE.java#L207-L210 will always trigger and throw the exception, even if we have other valid data formats, e.g. `ome.zarr`

constantinpape:
Fixes #805 and simplifies the checks for invalid storage types in general


constantinpape:
Thanks! Could you also push it on MoBIE-beta?

Issue #35, opened 2019-08-30
Add feature: see the version you are working with


vzinche:
I am not sure where to display it, but it would be nice to see the data version somewhere

tischi:
I added it to the window title of the new UI.

PR #906, opened 2022-11-16
Updates for revision

constantinpape:
Updating published projects and views


constantinpape:
This is adding all the updated figure panels now, as well as two new published projects.

For the figure panels there were two issues with the CLEM project I couldn't fix directly:
- https://github.com/mobie/clem-example-project/issues/35
- https://github.com/mobie/clem-example-project/issues/36

constantinpape:
All views are there now @tischi and this is good to be merged from my side.
(We can fix the issue in #35 independent of this)

PR #943, opened 2023-01-23
Issue942

tischi:


tischi:
- [ ] add views that combine the images and segmentations.

PR #969, opened 2023-02-23
Plateview

tischi:
The main additions are:

- OpenHCSDatasetCommand
- HCSDataSetter
- HCSPlate
- HCSPattern

I also changed a couple of things here and there.


constantinpape (commented):
This is too big for me to review in any detail, but overall this is great to have in MoBIE!
A few points:
- How do we actually use this? So how to open the tifs from a plate? See also https://github.com/mobie/mobie.github.io/pull/101#pullrequestreview-1312136626
- Does this also support segmentations?
- I have use for this ;) Should I send you some example data and a description of the naming scheme?

tischi:
> I have use for this ;) Should I send you some example data and a description of the naming scheme?

Great! Yes please, two wells are enough. Please open an issue about it.

> Does this also support segmentations?

Not yet; if you you also have example data please share along with a separate issue.

tischi:
Note to self: checks are failing because it depends on unreleased changes in mobie-io.

constantinpape:
@tischi see the first example data here: #971 
I will add an example with segmentation soon.

Issue #1013, opened 2023-04-13
overlay names for multiple sources is a problem


tischi:
When overlaying names for multiple sources in a grid this essentially does not work:

<img width="1482" alt="image" src="https://user-images.githubusercontent.com/2157566/231712612-e29d87e1-53b4-44db-845e-531936266022.png">

@imagejan Do you maybe have an idea how to conceptually solve this? I will write the code of course ;-)

tischi:
> Can the 'overlay names' be activated per layer? I guess in most cases the user would know they want to display the "wells" names, for example.

Thanks! That's probably an option....would either to mean another checkbox or we add it to the `[ S ]` (Settings button UI), for each layer. 

Also, maybe for grid views the behaviour could be different than in other cases.

FYI: here is another (beautiful imho) use-case:

- project: https://github.com/mobie/clem-example-project/
- view: SupplFig1a (zoomed in a bit)

<img width="904" alt="image" src="https://user-images.githubusercontent.com/2157566/233419750-340fe920-8289-41f1-a0d9-a9cd0015bea1.png">

To still enable this conveniently I think we would need to keep the current "overlay all names" option in addition to providing this for the individual layers?!


-------

> Related: how are the names generated currently? Taken directly from the table? Concatenated from the view name and the source?

Depends how the data is opened; I think is too complex to answer simply.

tischi:
I made a PR: https://github.com/mobie/mobie-viewer-fiji/pull/1063

Improvement ideas:

1. Draw black background below the texts (implemented in PR via [fb9817e](https://github.com/mobie/mobie-viewer-fiji/pull/1063/commits/fb9817ed97754cdaa9f27d0a1a69fc117b5ab2ed))
2. Collect all the text bounding boxes before drawing them and then only draw texts that do not overlap with any other text (TODO)

tischi:
I made a PR: https://github.com/mobie/mobie-viewer-fiji/pull/1063

Improvement ideas:

1. Draw black background below the texts (implemented in PR via [fb9817e](https://github.com/mobie/mobie-viewer-fiji/pull/1063/commits/fb9817ed97754cdaa9f27d0a1a69fc117b5ab2ed))
2. Collect all the text bounding boxes before drawing them and then only draw texts that do not overlap with any other text (TODO)

Issue #924, opened 2022-12-21
Specification for color string


tischi:
@constantinpape @K-Meech 

I am confused, in different places in the Java code I find different conventions how to encode/decode a color string:

- `(r=255,g=255,b=255,a=255)`
- `r=255,g=255,b=255,a=255`
- `255-255-255-255`

Could you please check what we actually say in the spec? 

For images and for values in tables?

- [ ] #934

constantinpape:
We have two different color specifications in the spec as well:
- `selectionColor`: `255-255-255-255` (see https://github.com/mobie/mobie.github.io/blob/master/schema/view.schema.json#L117)
- `color`(for the image display): `r=255,g=255,b=255,a=255`. (see https://github.com/mobie/mobie.github.io/blob/master/schema/view.schema.json#L132)

I agree that it would be good to unify this. My vote would be for `r=255,g=255,b=255,a=255` because it's more explicit.

tischi:
The issue with `r=255,g=255,b=255,a=255` is that I'd be very afraid that it may break a CSV table..
We could specify that it MUST be enclosed in quotation marks, but I am not sure this will always work, depending on the CSV parser....
In fact, did you also find the spec for the table? We support a color column...


maybe: `r255-g255-b255-a255` ?

tischi:
OK,  do you agree that commas are risky?

Could you find what we write in the spec for the colours in a table column?

Should I then change this in the places where we _write_ colors?

And I think for _reading_ colors I would support everything, i.e. try all conventions using a pattern matcher. Like this we would stay backwards compatible.

constantinpape:
> `r255-g255-b255-a255`

Ok, so we decided we will use this as the default encoding for colors (so use it when we write data, and in any description in the spec). We will support the old encodings for backwards compatibility.

I will update the spec tomorrow to support the new encoding.

constantinpape:
I updated this in the spec now.

tischi:
I forgot to implement this in the Java code until today, but now it should work.

tischi:
@constantinpape I am confused again

constantinpape:
@tischi you are correct, we want to use `r255-g255-b255-a255` as the default color string.
(I have misread the spec there, and we actually still support the other color strings as well.)

PR #1028, opened 2023-04-21
Json Test

tischi:
@constantinpape 

I cleaned up the code a bit.

I am confused now what actually the difference between the two validation methods is; in particular whether they us different libraries at all...would be cool if you could have a look!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -239,13 +239,7 @@
         <dependency>
             <groupId>com.github.erosb</groupId>
             <artifactId>everit-json-schema</artifactId>
-            <version>1.14.1</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.github.fge</groupId>
-            <artifactId>json-schema-validator</artifactId>
-            <version>2.2.14</version>
+            <version>1.14.2</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -242,5 +242,11 @@
             <version>1.14.1</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.fge</groupId>
+            <artifactId>json-schema-validator</artifactId>
+            <version>2.2.14</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/src/main/java/org/embl/mobie/lib/color/ColorHelper.java
+++ b/src/main/java/org/embl/mobie/lib/color/ColorHelper.java
@@ -68,8 +68,7 @@ public abstract class ColorHelper
 		if ( argbType == null ) return null;
 
 		final int colorIndex = argbType.get();
-		// FIXME: https://github.com/mobie/mobie-viewer-fiji/issues/924
-		final String string = "" + ARGBType.red( colorIndex ) + "-" + ARGBType.green( colorIndex ) + "-" + ARGBType.blue( colorIndex ) + "-" + ARGBType.alpha( colorIndex );
+		final String string = "r=" + ARGBType.red( colorIndex ) + ",g=" + ARGBType.green( colorIndex ) + ",b=" + ARGBType.blue( colorIndex ) + ",a=" + ARGBType.alpha( colorIndex );
 		return string;
 	}
 
--- a/src/main/java/org/embl/mobie/lib/color/ColorHelper.java
+++ b/src/main/java/org/embl/mobie/lib/color/ColorHelper.java
@@ -63,12 +63,14 @@ public abstract class ColorHelper
 		return getString( getARGBType( lut ) );
 	}
 
+	// convert a color to a string
+	// https://github.com/mobie/mobie-viewer-fiji/issues/924
 	public static String getString( ARGBType argbType )
 	{
 		if ( argbType == null ) return null;
 
 		final int colorIndex = argbType.get();
-		final String string = "r=" + ARGBType.red( colorIndex ) + ",g=" + ARGBType.green( colorIndex ) + ",b=" + ARGBType.blue( colorIndex ) + ",a=" + ARGBType.alpha( colorIndex );
+		final String string = "r" + ARGBType.red( colorIndex ) + "-g" + ARGBType.green( colorIndex ) + "-b" + ARGBType.blue( colorIndex ) + "-a" + ARGBType.alpha( colorIndex );
 		return string;
 	}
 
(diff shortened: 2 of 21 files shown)


constantinpape:
Hey @tischi,
do you really want to merge this into `main`? Wouldn't it make more sense to make this PR onto `develop`?
(And I am happy to take a look later)

tischi:
thanks for spotting this. changed to develop.

constantinpape:
> I am confused now what actually the difference between the two validation methods is; in particular whether they us different libraries at all...would be cool if you could have a look!

So, here's what I understand from some quick checks: there are three relevant jsonschema libraries:
1. https://github.com/everit-org/json-schema <- this is the one Kimberly has used, but we should NOT use it because it's deprecated (see their comment in the README)
2. https://github.com/java-json-tools/json-schema-validator <- this is the one proposed by ChatGPT's artificial stupidity and that we have now added
3. https://github.com/erosb/json-sKema  <- this is the library suggested in the readme of 1. (So we are not using it right now, but maybe we should).

I will see if I can reactivate my IntelliJ and have a look later.

constantinpape:
Hi @tischi,
I looked into this and we should use https://github.com/everit-org/json-schema after all. It supports what we need and the validation works correctly. (I think we did something wrong beforehand and that's why the spurious `name` didn't trigger the error).

I pushed a new commit now and the validation is working properly. And this shows that we still have the following spec related issues: (see full json that causes this below):
1. `"color": "255-255-255-255"`: this must now be `r=255,g=255,b=255,a=255`
2. `"overlayNames":false`: this is not part of the spec

Regarding 1: I am not sure where this is validated, can you fix it?
Regarding 2: I suggest we make this transient for now so that it's not serialized. Let's resolve #1013 first before discussing how we add this to the spec (because maybe we'll need to add some extra stuff for this).

```json
{"sources":{"testImage":{"image":{"imageData":{"bdv.n5":{"relativePath":"images/bdv-n5/testImage.xml","channel":0}}}}},"is2D":false,"defaultLocation":{"position":[0,0,0],"timepoint":0},"views":{"default":{"isExclusive":true,"sourceDisplays":[{"imageDisplay":{"showImagesIn3d":false,"visible":true,"sources":["testImage"],"color":"255-255-255-255","contrastLimits":[0,255],"name":"testImage","opacity":1}}],"overlayNames":false,"uiSelectionGroup":"bookmark"},"testImage":{"isExclusive":false,"sourceDisplays":[{"imageDisplay":{"showImagesIn3d":false,"visible":true,"sources":["testImage"],"color":"255-255-255-255","contrastLimits":[0,255],"name":"testImage","opacity":1}}],"overlayNames":false,"uiSelectionGroup":"testGroup"}}}
```

tischi:
1. see https://github.com/mobie/mobie-viewer-fiji/issues/924
2. changed to transient

constantinpape:
Can you push the changes here? The tests should pass then. (If not I will take another look)

tischi:
I changed it, but it still fails:

```
Json Schema validation failed:
#/sources/testImage: #: only 1 subschema matches out of 2
#/sources/testImage: #: 0 subschemas matched instead of one
```

I don't know what that means...

constantinpape:
@tischi: there is one problem left now:
It serializes this for the image source: `"testImage":{"image":{"imageData":{"bdv.n5":{"relativePath":"images/bdv-n5/testImage.xml","channel":0}}}}`.
However, we don't have channel in the spec for `bdv.n5`, only for `ome.zarr`.

So we either:
- do not serialize channel for `bdv.n5`
- add it to the spec for `bdv.n5` (for all sources)?

As far as I know the reason that we only have it for `ome.zarr` is that this is the only image format where we actually support multiple channels.

tischi:
I would say let's add it for all image sources to the spec.

from the Java code of `StorageLocation`:

```
	// the image data model in MoBIE is single channel,
	// and thus one needs to subset a potentially
	// multi-channel image data source.
	//
	// as data is opened as {@code SpimData}
	// the {@code channel} can also refer to a setup, which
	// does not need to be a fluorescence channel, but could be some
	// entirely other image, e.g. in CZI or LIF files
	// 
	// also the BDV XML based files can contain multiple "setups"
	// which can be chosen from via the {@code channel} field
	//
	// historically, we only had fluorescence channels, 
	// and thus the name of the variable is "channel" and cannot 
	// readily be changed because we use in the MoBIE JSON spec
	public Integer channel = 0; // only for images
```

constantinpape:
Hey @tischi,
I have updated the spec now and the image tests pass. However the segmentation tests still fail because of this:
```"tableData":{"tsv":{"relativePath":"tables/testImage","channel":0}
```
I.e. the `channel` attribute is also serialized for `tableData`. This doesn't make sense and you will need to fix it.

constantinpape:
> That's right, and I somehow always thought the reason I am doing this is that I need to do this because of the spec. I could change this, and it would be nice but a lot of refactoring.
> 
>     * [ ]  I think we should open a new issue about changing this in the code

Your change works for now (so indeed looks like setting it to `null` will not serialize it), so it's up to you if you want to refactor it.

> However for me the tests still fail.

The remaining problem is that the `segmentationDisplay` serializes the `blendingMode` property. In the spec we only have this for `imageDisplay`. I think this goes back to a discussion some time ago where we said that the blending mode is not needed for segmentations because we only have one reasonable choice for it. So again two ways to fix it:
- don't serialize the `blendingMode` for segmentation displays
- or we add it to the spec for segmentation displays (but I would prefer to do this only if it actually makes sense to have it there)

> Can you somehow change the code such that it prints something as to why the tests fail?

Sorry, I don't really know how to do this. The problem is that the errors from the java jsonschema checker are somehow nested, and [this](https://github.com/mobie/mobie-viewer-fiji/blob/jsontest/src/test/java/org/embl/mobie/lib/create/JSONValidator.java#L57-L59) only prints the first level (so we only see the unhelpful error messages). I am not sure how to change it to print this recursively. 

For now, I have been copying the json representation from debug and pasted it into a script to validate it via the python jsonschema library to get the proper errors.

tischi:
I did the same trick of removing the default value, does it work?

I think here I could also move the field to just `ImageDisplay` but I am not sure we may need it also for the `AnnotationDisplay` at some point, so if possible I would leave it also there....

constantinpape:
Your changes didn't work, the `blendingMode` is still serialized (and consequently the tests still fail).

tischi:
now?

constantinpape:
> now?

Yes, now it works; the tests pass.

This is good to go from my end now.

martinschorb:
Hi,

I just realized that also MoBIE's "Save Current View" functionality has updated from putting "name" fields (violating the current spec) into views. However, it now puts adds another field that does not validate correctly:`Property 'overlayNames' is not allowed`

It is doing that even when I remove all of these properties from the source dataset. And it then adds it to all views of the dataset.

Like here:
https://github.com/mobie/centriole-tomo-datasets/blob/9041c0f52424a5a257101941afd10f5e13153396/data/tomo/dataset.json#L50 

Or am I using an outdated spec (current `master`)?

tischi:
@martinschorb I will merge this now and upload to MoBIE-beta later today. Then you can check again. If the issue persists, please open an issue.

PR #1040, opened 2023-05-23
Add contributing guidelines

k-dominik:
I am aware that development setup is also up to taste - I summarized here what worked for me. I think having _something_ here is better than nothing.

Summary:

* added initial contributing guidelines to ease onboarding of new developers
* cleanup: removed scripts generated by `install.sh`
* Note: in `install.sh` I removed the custom `JAVA_HOME` - when conda for development (and maybe also in general?!) this should not be necessary to set manually


tischi:
Thanks a lot! Also this is awesome!!

Shall we add something to CONTRIBUTING.md that this is also shipped as a Fiji plugin?
Right now it feels a bit conda CLI centric...

k-dominik:
Well, if there is a way to run fiji with this development version of the plugin, yes, then this should definitely be documented. I have no clue how to do that, so could you maybe add this?

PR #1041, opened 2023-05-23
added tooltips to some of the user interface

k-dominik:
I added some tooltips to the tiny buttons in the UI that should help users to make a _little_ more sense of their function.
If you have any better ideas for tooltips I'm happy to modify them!

Cheers


tischi:
This is great! Thank you! I was wondering whether we could in addition, instead of `C` and `B` replace the button content with some little icons such as

image

image

Do you think that could be worth looking into?

k-dominik:
icons would be even nicer :)

PR #1046, opened 2023-06-28
Issue1044

tischi:


tischi:
@imagejan after the above fixes, using `Load Columns...` it looks now like this:

I would say this is correct, you too?

image

imagejan:
@tischi I can confirm this now works as expected in the _Open Image and Labels..._ case. 🚀

Unfortunately, it still doesn't in our MoBIE project where we added the segmentation as a merged grid view with multiple sources (from 240 wells). I still get:

> There are duplicate columns: [label]
> Those columns will be replaced by the columns in the newly loaded table.

... and all the labels turn `0`.

If you don't have an immediate idea what might cause this, I'll try to set up a minimal example project to reproduce the issue.

tischi:
@imagejan this is strange indeed.

I guess `label` is the name of the label id column, correct?

Then those [two lines of code](https://github.com/mobie/mobie-viewer-fiji/blob/b628e48e0bc68c8821ece91d2d11da0b2b1e623d/src/main/java/org/embl/mobie/lib/table/saw/TableSawAnnotationTableModel.java#L186):

```Java
final List< String > additionalColumnNames = additionalTable.columnNames().stream().filter( col -> ! mergeByColumnNames.contains( col ) ).collect( Collectors.toList() );
final List< String > duplicateColumnNames = additionalColumnNames.stream().filter( col -> columnNames.contains( col ) ).collect( Collectors.toList() );
```

...should be such that `duplicateColumnNames` **cannot** contain `label`, because it _should_ be part of `mergeByColumnNames`. 

So, if you run in debug mode and set a breakpoint there and check what is going on, would probably allow us to figure out the issue.

imagejan:
@tischi I now uploaded an example dataset (ome-zarr and MoBIE) illustrating the issue here:
https://doi.org/10.5281/zenodo.8141917

Please let me know if you can reproduce.

tischi:
@imagejan 

There are two segmentations: `cellpose` and `cellpose_1`, which one should I use?

image

And, to make sure you are going:  `Table > Load Columns` ...  Load from `Project` ...  `additional.tsv`, right?

imagejan:
@tischi wrote:
> `Table > Load Columns ...` Load from `Project` ... `additional.tsv`, right?

Yes, correct.

> cellpose and cellpose_1, which one should I use?

This was just to illustrate segmentations from two different channels in the same labels group in the zarr file, and how we can share the same common table source for both of these segmentation sources.

It shouldn't matter for the 'Load Columns' step, as the existing label ids are the same in both channels.

tischi:
@imagejan I think I found several bugs. 

For me it works now for the example data that you attached. Could you please build from this PR (`branch1044`) and confirm?

When loading the columns one gets errors like this:

`Exception in thread "AWT-EventQueue-0" java.lang.ArrayIndexOutOfBoundsException: 10 >= 10
	at java.util.Vector.elementAt(Vector.java:479)`

This is something I have been fighting with forever and still don't understand it, but I think you can simply ignore it.....

imagejan:
@tischi great, thanks a lot!

I confirm that the `Load Columns...` step now works as expected 👍 🎉

PR #1062, opened 2023-11-19
Improve ColorByColumn GUI

tischi:
@constantinpape @martinschorb 

Currently we are supporting these LUTs:

```
GLASBEY,
RGBA_COLUMN,
BLUE_WHITE_RED,
VIRIDIS,
```

I think the first two only work for categorial data (i.e. columns of type `String.class`), while I think the last two only work for numerical data. Is that true (sitting in a train with bad internet, thus cannot test right now)?

If it is true I think it may be better to make the color by column GUI a two stage process where

GUI 1: only select the column
GUI 2: only present **_valid_** choices for the LUTs, depending on whether or not `table.columnClass( columnName ) instanceof String`

This would avoid that users can pick invalid combinations.

What do you think?

### DIFF
--- a/src/main/java/org/embl/mobie/MoBIESettings.java
+++ b/src/main/java/org/embl/mobie/MoBIESettings.java
@@ -28,14 +28,11 @@
  */
 package org.embl.mobie;
 
-import org.embl.mobie.io.ImageDataFormat;
 import org.embl.mobie.lib.io.DataFormats;
 import org.embl.mobie.lib.serialize.View;
 import org.embl.mobie.lib.table.TableDataFormat;
 
-import java.util.ArrayList;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 
 public class MoBIESettings
--- a/src/main/java/org/embl/mobie/lib/SourcesFromTableCreator.java
+++ b/src/main/java/org/embl/mobie/lib/SourcesFromTableCreator.java
@@ -71,9 +71,9 @@ public class SourcesFromTableCreator
 				String fileName = table.getString( 0, imageColumn);
 				String relativeFolder = table.getString( 0, imageColumn.replace(  "FileName_", "PathName_" ) );
 				String referenceImagePath = MoBIEHelper.createAbsolutePath( root, fileName, relativeFolder );
-				IJ.log( "Determining number of channels of " + imageColumn + ", using " + referenceImagePath + "..." );
+				IJ.log( "Determining number of channels of " + imageColumn + " from " + referenceImagePath + "..." );
 				int numChannels = MoBIEHelper.getMetadataFromImageFile( referenceImagePath, 0 ).numChannelsContainer;
-				IJ.log( "Number of channels is " + numChannels );
+				IJ.log( "Number of channels: " + numChannels );
 				for ( int channelIndex = 0; channelIndex < numChannels; channelIndex++ )
 				{
 					imageFileSources.add( new ImageFileSources( imageColumn + "_C" + channelIndex, table, imageColumn, channelIndex, root, gridType ) );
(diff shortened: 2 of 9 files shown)


constantinpape:
Overall I think this is a good idea. One comment:

>  think the first two only work for categorical data (i.e. columns of type String.class)

In theory I think these could also work for int datatypes since those can also encode categorical rather than numerical data. I am not sure how exactly it is implemented and if we want to allow this. But if you touch that again it would be a good idea to check what happens for int data (is there a real distinction between int and float in the tables?) and what we think should happen.

tischi:
Yes, I agree @constantinpape, I had a look at the code and the colours for categorical data are assigned like this:

```
	final Object value = input.getValue( columnName );
	if ( value == null )
		output.setZero();
	else
		convertStringToARGB( value.toString(), output );
```

Thus, I guess it should already work for anything? Could you try what happens if you choose a numeric column and color it with Glasbey? Would try myself, but I am at a conference with bad internet...

tischi:
@constantinpape @martinschorb 

It turns out that:

1. One could also use Glasbey for numerical columns
2. One could also use BlueWhiteRed for categorical columns

While this may not always make sense I changed the code now to enable this. 

This results in:

1. Keeping the UI as is
2. Preventing crashes when choosing BlueWhiteRed for Categorical columns
3. Enabling all column class LUT combinations

The key change is to now simply:

```
		if ( Number.class.isAssignableFrom( table.columnClass( columnName ) ) )
		{
			final Pair< Double, Double > minMax = table.getMinMax( columnName );
			return ColoringModels.createNumericModel( columnName, lut, minMax, true );
		}
		else
		{
			return ColoringModels.createCategoricalModel( columnName, lut, TRANSPARENT );
		}
```

------

I realised that this probably is now inconsistent with the JSON spec, which I think relies on this logic:

```
	public static boolean isNumeric( String lut )
	{
		return lut.contains( BLUE_WHITE_RED ) || lut.contains( VIRIDIS );
	}

	public static boolean isCategorical( String lut )
	{
		return lut.contains( GLASBEY ) || lut.contains( ARGB_COLUMN ) || lut.contains( RGBA_COLUMN );
	}
```

That means, what can happen now is:

1. User configures blue-white-red for a non-numeric column and saves the current view to JSON
2. Reloading that view would find that blue-white-red is a "numeric LUT" and thus assume that the column also is numeric and it will **crash**. In fact, it **crashes** already here: `annotationDisplay.getValueLimits()`, because it cannot find the required value limits for the "numeric LUT"

Since I found now that all LUTs can be applied to all data I think we should change the logic and _not_ make the LUTs numeric or categorical but rather the table columns.

That would probably mean adding another field to the JSON that indicates whether the coloured column is numeric or not:

```
            "lut": "blueWhiteRed",
            "colorByColumn": "FileName_HighZoom.Focus_IMG",
```

Or, as I just realised, probably it is fine to check wether `valueLimits != null`

```
            "lut": "glasbey",
            "colorByColumn": "Rotation_NUM",
            "valueLimits": [
              -60.29256771,
              63.6254595
            ],
```

because whenever a numerical colouring model is serialised there MUST be `valueLimits`?! 
@constantinpape can you confirm? what do you think? I think it would be great to get rid of the concept of numeric LUTs.

tischi:
I found in the spec:

> lut: The look-up-table for categorical coloring modes. Note: if the lut is numeric (‘viridis’, ‘blueWhiteRed’), the valueLimits field MUST be given. ‘argbColumn’ is deprecated, please use ‘rgbaColumn’ instead.

constantinpape:
Sorry, I didn't have time to go through everything here. But from what I understand everything works out with the current version of the spec as you can just check whether `valueLimits` is given or not.

PR #1064, opened 2023-11-22
Read OME-Zarr HCS from S3

tischi:


tischi:
- [ ] the site ids start at 0 and the site position code thinks they should start at 1

tischi:
Should be fine to merge as is.

Issue #1116, opened 2024-03-13
Link to existing OME-Zarr images in Fiji project creator


tischi:
@K-Meech 

You had:

```
            gd.addMessage( "Note: You can only 'link' to images outside the project folder \n" +
                    " for local projects. 'copy' or 'move' if you wish to upload to s3" );
```

I don't see why one could not link to OME-Zarr from and to S3? In fact, I would think in the future it could be a major use-case to link to OME-Zarr hosted on S3 at the BioImage Archive.

tischi:
Same question here:

```
            if ( imageDataFormat == ImageDataFormat.OmeZarr && addMethod == ProjectCreator.AddMethod.link )
            {
                IJ.log( "link is currently unsupported for ome-zarr. Please choose copy or move instead for this file format." );
                return;
            }
```

Why is that? We could just store the absolutePath to the linked OME-Zarr; I don't see the problem....

K-Meech:
> ```
>             gd.addMessage( "Note: You can only 'link' to images outside the project folder \n" +
>                     " for local projects. 'copy' or 'move' if you wish to upload to s3" );
> ```

This message was about people linking to images stored on the filesystem outside of their project. It's useful to be able to do this so people don't have to copy large images around when they just want to view their mobie project locally. The issue is that when this project is then uploaded to S3, these links will break and the images will no longer be available. 

I agree - it should be possible to link to ome-zarr images that are already hosted on S3. This just hasn't been implemented yet, and would need to be some kind of new option.
```
            if ( imageDataFormat == ImageDataFormat.OmeZarr && addMethod == ProjectCreator.AddMethod.link )
            {
                IJ.log( "link is currently unsupported for ome-zarr. Please choose copy or move instead for this file format." );
                return;
            }
```

For this one, it's just that this hasn't been implemented yet. So far all linking has been done by placing an XML file inside the MoBIE project that internally contains a link to the position of the file elsewhere on the filesystem. I agree that using absolute path is probably a better way to do this - and would mean that n5 + ome-zarr could be handled in the same way.

tischi:
@K-Meech would you have some time to look into the `new-n5` branch to see how we could add support linking to existing OME-Zarr datasets in the ProjectCreator via the absolute path mechanism?

K-Meech:
@tischi I probably wouldn't have time until next week or the week after. Do you need it done before then?

tischi:
[...]

PR #1129, opened 2024-04-28
Fix linking

K-Meech:
For https://github.com/mobie/mobie-viewer-fiji/issues/1116

This PR:
- Fixes linking images within the project creator (uses absolutePath if the image is outside the project directory)
- Fixes remote metadata creation
- Fixes failing project creator tests + adds more for linking / remote metadata
- Updates the javadoc

Note: I've currently commented out the line in `ImagesCreatorTest` that validates the dataset json against the schema. This is due to the missing `absolutePath`. It can be un-commented once this is added.

Also, this doesn't currently support directly giving an `s3address` to the project creator (for files that are already stored remotely). This is probably best added later, as it's a new feature that wasn't in the original project creator.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,7 @@
         <n5-zarr.version>1.3.1</n5-zarr.version>
         <n5-hdf5.version>2.2.0</n5-hdf5.version>
         <n5-ij.version>4.1.1</n5-ij.version>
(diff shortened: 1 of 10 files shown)


K-Meech:
Here's a link to the PR to add `absolutePath` into the schema: https://github.com/mobie/mobie.github.io/pull/110

PR #1183, opened 2024-10-23
fix: correct path resolution with root directory handling

ksugar:
- Fixed the logic for resolving paths by using the root directory when it's not null.
- If the root directory is null, the path is now resolved using the Cellprofiler column.


tischi:
Thanks @ksugar ! 
I am impressed that you dug so deep into the code in such a short time !

Issue #1148, opened 2024-06-12
Delete view from UI


tischi:
There are several user requests that it would be great to enable _deleting_ a view from the UI.

@K-Meech is that something that you'd like to have a look at?

K-Meech:
@tischi the next few weeks are super busy for me, so it would be a while before I could look at this. If you have time sooner - then go ahead! Otherwise I can look at it in a few weeks time.

K-Meech:
@fstur unfortunately I haven't had the chance to implement this yet. I should have some time this week / next week to look into adding this though!

tischi:
Hi @fstur,

We released a new version where this feature should be present. 

Could you please test it?

To access it: Right click in the BDV window and select "Delete view" from the drop down.

tischi:
Nice!

Thanks @K-Meech !!

PR #1193, opened 2024-12-01
Allow deletion of views via the UI

K-Meech:
For https://github.com/mobie/mobie-viewer-fiji/issues/1148

This PR allows deletion of views via right clicking in the MoBIE viewer and selecting `Delete View`. Similar to the view saving options, it allows deleting views from the 'current project' (i.e. the dataset.json) or an 'external file' (any view .json saved on the file system). The 'current project' option is currently only implemented for local projects - if you try to use it with a github / s3 project it should throw an error.

Let me know if any changes are needed! I tried it out locally with a few example projects, but it would be great if more people could test this out to check everything is working correctly :smile:

### DIFF
--- a/src/main/java/org/embl/mobie/lib/bdv/view/SliceViewer.java
+++ b/src/main/java/org/embl/mobie/lib/bdv/view/SliceViewer.java
@@ -70,6 +70,7 @@ public class SliceViewer
 	public static final String UNDO_SEGMENT_SELECTIONS = "Undo Segment Selections [ Ctrl Shift N ]";
 	public static final String LOAD_ADDITIONAL_VIEWS = "Load Additional Views";
 	public static final String SAVE_CURRENT_SETTINGS_AS_VIEW = "Save Current View";
+	public static final String DELETE_VIEW = "Delete View";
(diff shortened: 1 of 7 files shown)


tischi:
Hi @K-Meech, I don't manage to locally check out your PR.
Would you mind doing this PR from a branch _within this repo_ rather than from your fork?

K-Meech:
I think you can checkout the PR locally with:
```
git fetch origin pull/1193/head:test-pr-1193
git checkout test-pr-1193
```
Does that work?

tischi:
Yes, thanks!

tischi:
@K-Meech I made a branch `test-pr-1193` where I added an error message if the dataset.json is not found. The issue is that there are now many other ways of opening MoBIE that do not include a dataset.json. 
The clean solution would be to not even show the "Delete from current project" option.
I was a bit in a rush right now. I think I would still try to implement this rather than throwing the error...

tischi:
Hi @K-Meech,

Shall we just merge? Including my additions (see above)?

K-Meech:
Sounds good to me - go ahead @tischi !

PR #1227, opened 2025-03-28
Support S3 omeZarr in project creator

tischi:
This is WIP, I just started...

### DIFF
--- a/src/main/java/org/embl/mobie/lib/create/ImagesCreator.java
+++ b/src/main/java/org/embl/mobie/lib/create/ImagesCreator.java
@@ -296,23 +296,25 @@ public class ImagesCreator {
                 imageDataFormat,
                 ThreadHelper.sharedQueue );
 
-        if ( ! isImageValid( imageData, projectCreator.getVoxelUnit() ) ) {
-            return;
-        }
-
         if ( ! is2D( imageData ) && projectCreator.getDataset( datasetName ).is2D() ) {
             throw new UnsupportedOperationException("Can't add a 3D image to a 2D dataset" );
         }
 
         // If an image of the same name is inside the project, delete it when overwrite=True.
-        File oldImageFile = new File(imagesDirectory, imageName + ".ome.zarr" );
-        if (oldImageFile.exists()) {
-            if (overwrite) {
-                IJ.log("Overwriting image " + imageName + " in dataset " + datasetName );
-                deleteImageFiles( datasetName, imageName );
-            } else {
-                throw new UnsupportedOperationException("An image called " + imageName + "already exists in the dataset " +
-                        datasetName + ", and overwrite is set to false" );
+        if ( addMethod.equals( ProjectCreator.AddMethod.Copy ) )
+        {
+            File oldImageFile = new File( imagesDirectory, imageName + ".ome.zarr" );
+            if ( oldImageFile.exists() )
+            {
+                if ( overwrite )
+                {
+                    IJ.log( "Overwriting image " + imageName + " in dataset " + datasetName );
+                    deleteImageFiles( datasetName, imageName );
+                } else
+                {
+                    throw new UnsupportedOperationException( "An image called " + imageName + "already exists in the dataset " +
+                            datasetName + ", and overwrite is set to false" );
+                }
             }
         }
 
@@ -320,7 +322,7 @@ public class ImagesCreator {
             projectCreator.setVoxelUnit( imageData.getSourcePair( 0 ).getB().getVoxelDimensions().unit() );
         }
 
-        File newImageFile;
+        File newImageFile; // FIXME: This should be a URI String to also work with S3
         switch (addMethod) {
             case Link:
                 // Do nothing, the relative or absolute path to the linked image will be added to the dataset.json
@@ -328,6 +330,8 @@ public class ImagesCreator {
                 break;
             case Copy:
                 newImageFile = copyImage( uri, imagesDirectory, imageName);
+                break;
+            case LinkS3:
                 break;
             default:
                 throw new IllegalStateException("Unexpected value: " + addMethod);
--- a/src/main/java/org/embl/mobie/lib/create/ui/ProjectsCreatorUI.java
+++ b/src/main/java/org/embl/mobie/lib/create/ui/ProjectsCreatorUI.java
@@ -447,6 +447,7 @@ public class ProjectsCreatorUI extends JFrame {
             final GenericDialog gd = new GenericDialog( "Add OME-Zarr To Project..." );
             String[] addMethods = new String[]{
                     ProjectCreator.AddMethod.Link.toString(),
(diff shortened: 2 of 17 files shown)


tischi:
Hi @K-Meech, could you have a look whether this looks reasonable to you and whether it could be merged? 

The main new feature is that one can also link to external S3 hosted OME-Zarr.

I realised that we could support also linking to multi-channel images, presenting the user a UI to select one, but this could be done in a separate PR. I added a bunch of TODO at places where we have to work on.

K-Meech:
Thanks @tischi - this week is really busy, so I probably can't get to this until Saturday at the earliest, or maybe even next week. Is that fine?

tischi:
Next week is fine! Thanks!

K-Meech:
@tischi - do you have an example remote ome-zarr image I could try this out with?

tischi:
Sure: `https://s3.embl.de/i2k-2020/platy-raw.ome.zarr` :-)

tischi:
In fact:

- https://s3.embl.de/i2k-2020/platy-raw.ome.zarr  
  - This may throw you an error, because it contains multiple datasets
- https://s3.embl.de/i2k-2020/platy-raw.ome.zarr/labels/cells
  - This  should work though

src/main/java/org/embl/mobie/lib/create/ImagesCreator.java
```diff
@@ -296,48 +296,47 @@ public void addOMEZarrImage( String uri,
...
         // If an image of the same name is inside the project, delete it when overwrite=True.
-        File oldImageFile = new File(imagesDirectory, imageName + ".ome.zarr" );
-        if (oldImageFile.exists()) {
-            if (overwrite) {
-                IJ.log("Overwriting image " + imageName + " in dataset " + datasetName );
-                deleteImageFiles( datasetName, imageName );
-            } else {
-                throw new UnsupportedOperationException("An image called " + imageName + "already exists in the dataset " +
-                        datasetName + ", and overwrite is set to false" );
+        if ( addMethod.equals( ProjectCreator.AddMethod.Copy ) )
```
K-Meech, line 304: Don't you want this for linking too? Say there's a local ome-zarr image stored in your MoBIE project named "test". If I now add a link to a remote image called "test" - would this still be an issue? I'd assume you only want one image per name.

src/main/java/org/embl/mobie/lib/create/RemoteMetadataCreator.java
```diff
@@ -112,11 +108,11 @@ private void addRemoteMetadataForImage( String datasetName, String imageName ) t
                         datasetName,
```
K-Meech, line 108: Above (line 95), only `relativePath` are allowed when projects are uploaded to S3 (to prevent local links to images outside of the MoBIE project from being broken on upload). I guess an absolute path that was to S3 would work though - do you want to allow this case?

src/main/java/org/embl/mobie/lib/create/ImagesCreator.java
```diff
@@ -296,48 +296,47 @@ public void addOMEZarrImage( String uri,
                 imageDataFormat,
                 ThreadHelper.sharedQueue );
 
-        if ( ! isImageValid( imageData, projectCreator.getVoxelUnit() ) ) {
```
K-Meech, line 299: It would be good to keep this check if possible? It ensures that the image units remains consistent throughout the entire project + rejects multi-channel images. Maybe keep it until you've added an extra menu to allow the user to select a specific channel?

src/main/java/org/embl/mobie/lib/create/ui/ProjectsCreatorUI.java
```diff
@@ -461,18 +461,20 @@ public void addOMEZarrDialog() {
             exclusive = gd.getNextBoolean();
             useFileNameAsImageName = gd.getNextBoolean();
 
-            String filePath = getOMEZarrImagePathDialog();
-
-            if ( filePath != null )
-            {
-                addOMEZarr( filePath, datasetName );
-            }
+            // TODO: a Browse button for local folders would be nice
```
K-Meech, line 464: I agree! If possible, it would be great to keep the file browser for finding local ome-zarr images. It's much harder to find this path manually to paste in.

src/test/java/org/embl/mobie/lib/create/ImagesCreatorTest.java
```diff
@@ -195,7 +195,7 @@ Object copyImageIntoDataset(boolean is2D ) {
         // NB: Avoid recursive class loading of ij.* classes.
         ImagePlus image = (ImagePlus) writeImageOutsideProject( is2D );
```
K-Meech, line 197: Ideally, it would be great to add new tests to cover linking / copying etc for remote images. It should make the project creator easier to maintain - although, with the tests github action currently failing it's hard to spot when changes are breaking these tests.

src/main/java/org/embl/mobie/lib/create/RemoteMetadataCreator.java
tischi: Yes, I think we should allow the absolute path linking to S3.

src/main/java/org/embl/mobie/lib/create/ImagesCreator.java
tischi: Yes, only one image per name.

src/main/java/org/embl/mobie/lib/create/ui/ProjectsCreatorUI.java
tischi: Yes, but we can do this in the future.

src/test/java/org/embl/mobie/lib/create/ImagesCreatorTest.java
tischi: It would be great to add some tests!
Could you?
At least locally the tests are running and thus useful.

K-Meech:
Sorry for the delay in getting back to you @tischi - all sounds good to me! Are you happy to make those changes on this branch? I can look at adding some tests, but it probably won't be until near the end of May (this month is very full 😅 )

tischi:
Hi @K-Meech,
Unfortunately, I am also quite busy.
I would lean towards waiting until you find some time to implement the needed changes.

tischi:
Hi @K-Meech, any chance you could work on this at some point?

K-Meech:
Hi @tischi - sorry for the delay in getting back to you on this 😅 I'll try to find time to look at this either this weekend, or the one after.

Taking a quick look over the changes here - I noticed there are many that seem un-related to the project creator? For example, changes to `BoxSelectionCommand`, `MaskImagesCommand` etc... If you had time, could you possibly fix the merge conflicts on this branch + remove changes that are un-related to the project creator? This would make it a lot easier for me to review / make fixes and so on. Thanks!

tischi:
@K-Meech done.

PR #1266, opened 2025-12-03
Integration of BigVolumeBrowser

ekatrukha:
This PR adds integration of [BVB](https://github.com/UU-cellbiology/bigvolumebrowser) for large 3D volumes / segmentations/  spots (replaces old [bvv-playground](https://github.com/UU-cellbiology/bvv-playground) usage).

Now added sources (especially segmentations) can be rendered as surfaces with lighting,   
clipping can be performed and etc. Here is an example (in the beginning old, "plain" rendering is shown)

https://github.com/user-attachments/assets/1ae00775-d5af-42bf-84b1-c5115bc3bbf1

Additionally, I've added `SpotDisplay` spots rendering using 3D shaders, which is fast and memory efficient, see below

https://github.com/user-attachments/assets/075b2348-7e01-4de9-b84e-1991f6691fdc

It is a proof of principle, so some UI functions/colors updates still need to be implemented.

+ I fixed some other things, like: 
- Volume visibility change inside BVB now is reflected in MOBiE interface;
- removed BVB/BVV settings from the context menu, since BVB has its own;
- removed `SaC` wrappers, they now also live in BVB.

@tischi please try it and let me know if you have comments or found some bugs.
It relies on some SNAPSHOTS, so it is still WIP.
But if everything looks ok, I can do releases, and we can merge this.

Also, we need to decide how to ship it.
Either ship "full BVB" or just mention in the installation to include BVB update site.
The last one is easier, but then any click on "B" checkbox will give a lot of errors.
I do not have strom opinions about it.

Cheers,
Eugene

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,15 @@
                 <role>support</role>
             </roles>
         </contributor>
+        <contributor>
+            <name>Eugene Katrukha</name>
+            <roles>
+                <role>developer</role>
+                <role>debugger</role>
+                <role>reviewer</role>
+                <role>support</role>
+            </roles>
+        </contributor>
     </contributors>
     <mailingLists>
         <mailingList>
@@ -87,8 +96,8 @@
         <bigdataviewer-playground.version>0.11.2</bigdataviewer-playground.version>
         
         <!-- BigVolumeBrowser -->
- 		<bigvolumebrowser.version>0.0.5-SNAPSHOT</bigvolumebrowser.version>
-        <bvv-playground.version>0.5.1-SNAPSHOT</bvv-playground.version>
+ 		<bigvolumebrowser.version>0.0.5</bigvolumebrowser.version>
+        <bvv-playground.version>0.5.1</bvv-playground.version>
 
         <annotations.version>16.0.2</annotations.version>
         <trove.version>1.0.2</trove.version>


tischi:
This is amazing! I love how you integrated everything into the BDV panels! 

Is it possible to render spots such that they get some 3-D shading appearance?

image

tischi:
When I am in the clipping mode and cannot rotate the view anymore, i.e. the BDV mouse navigation does not seem to work anymore.

image

And as mentioned, it would be nice to have some way to have the surfaces of labels appear less "grizzely":

image

In the above example there seem to be only three different shades of the colour, leading to very abrupt colour jumps.

ekatrukha:
For the points I can try to make some "fake lighting" mode, yes.

Can you please tell me how to reproduce the clipping "stuck" mode? What datasets do you use?

tischi:
> For the points I can try to make some "fake lighting" mode, yes.

Nice 

> Can you please tell me how to reproduce the clipping "stuck" mode? What datasets do you use?

1. Open [this collection table](https://docs.google.com/spreadsheets/d/1jEnl-0_pcOFQo8mm8SUtszoWewvjyFXY0icO7gPUaQk/edit?gid=0#gid=0)
2. Show the segmented nuclei layer in BVV
3. Activate the Clipping

-------

And any ideas for how to get less abrupt colour changes for the label rendering?

tischi:
> It was your advice to put everything in the cards panel, and I think it works better.

If I would start over with MoBIE I would also try putting everything in the cards panel...

PR #1269, opened 2026-01-13
Add Maven Scijava version badge to README

NicoKiaru:
Hi @tischi, 

I find these badges pretty convenient when you look for the current version. Here's one example if you wish. I usually make them with a llm, giving an example, the github address and the group and artifact (provided the repo is published in maven scijava)

Cheers!


tischi:
Thanks!

PR #1325, opened 2026-08-24
Add numericAnnotation data source type

cyrilcros:
Declare derived image sources in dataset.json that wrap an annotation (segmentation/spot) label image and map each pixel to a numeric value from a table column.


tischi:
Hi, thank you very much for adding this. That's a good point. This was missing.

I'm a little bit reluctant adding that because that would mean we really support it. 

As mentioned talking to you, I think we have now a better way of creating the same visualizations.

But I understand that you need it for the publication and it's not so much code so I think maybe we can maintain it.

cyrilcros:
Hi, sorry I should have pushed this again my current repo https://github.com/cyrilcros/mobie-viewer-fiji/pulls, that was a mistake...
I wanted to talk

PR #1326, opened 2026-08-25
Add numericAnnotation data source type

cyrilcros:
## Summary

Adds a new `numericAnnotation` data source type to `dataset.json` that dynamically generates an `Image<DoubleType>` from an annotation (segmentation/spot) label image plus a table column. Each pixel belonging to an annotation carries that annotation's numeric value for the column; background pixels are zero.

The resulting source behaves like any standard intensity image — color (LUT), contrast, invert, opacity, and blending all work identically through the existing `ImageDisplay`/`ImageSliceView` path.

## Changes

- **`NumericAnnotationDataSource`** (new, `lib/serialize/`): fields `annotationSource` + `column`.
- **`DataSourceMapAdapter`**: registers the `"numericAnnotation"` type key.
- **`MoBIE.initDataSource()`**: no-op branch (deferred handling — the base image isn't ready yet when sources load in parallel).
- **`ViewManager.initData()`**:
  - includes each `annotationSource` in the set of sources to initialise, so a view can reference a numeric annotation without also listing its base segmentation;
  - after all base sources load, wraps the base annotation image in `NumericAnnotationImage` and registers it via `DataStore.addImage`.

## Example

dataset.json:

```json
{
  "sources": {
    "nuclei": {
      "segmentation": {
        "imageData": { "0": { "relativePath": "images/nuclei" } },
        "tableData": { "tsv": { "relativePath": "tables/default.tsv" } }
      }
    },
    "nuclei-anchor_z": {
      "numericAnnotation": { "annotationSource": "nuclei", "column": "anchor_z" }
    }
  },
  "views": {
    "default": {
      "sourceDisplays": [
        { "segmentationDisplay": { "sources": ["nuclei"] } },
        { "imageDisplay": { "sources": ["nuclei-anchor_z"], "contrastLimits": [0, 280] } }
      ]
    }
  }
}
```

## Testing

Verified locally against a project with a nuclei segmentation and a table column (`anchor_z`): the numeric annotation layer loads as an image source, renders correctly, and supports recolor/contrast like any other intensity image.

### DIFF
--- a/src/main/java/org/embl/mobie/MoBIE.java
+++ b/src/main/java/org/embl/mobie/MoBIE.java
@@ -746,6 +746,13 @@ public class MoBIE
 			regionTableSource.table = TableOpener.open( tableLocation, tableFormat );
 			DataStore.addRawData( regionTableSource );
 		}
+		else if ( dataSource instanceof NumericAnnotationDataSource )
+		{
+			// NumericAnnotationDataSources are expanded after all base sources
+			// have been initialised (see ViewManager.initData()).
+			// The base annotation image must already be in DataStore
+			// before the NumericAnnotationImage wrapper can be created.
+		}
 
 		if ( log != null )
 			IJ.log( log + dataSource.getName() );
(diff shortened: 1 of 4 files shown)


cyrilcros:
I tried testing that locally (saving/loading), seems okay for now.

tischi:
Thank you @cyrilcros! 
I will merge it and then I can also make a release for the update site. Do you need that soon?

cyrilcros:
That would be nice if we could try a release before the end of the week, but I want to try to get my "MoBIEuniverse" feature merged first, it is also small. I am still testing it.
I will reach out.

By the way, do you have a prefered way of converting a for loop to something parallel? I am thinking about the nuclei rendering one by one, independently of trying to cache the renders.

tischi:
Hi, yes, you can try to use this for parallelization:

`ThreadHelper.executorService`

if you ask AI for help, I think you should figure it out.

PR #1328, opened 2026-09-04
Persistent disk-backed mesh cache for segment 3D rendering

cyrilcros:
Adds a persistent, disk-backed cache for the smoothed segment meshes shown in the Fiji 3D viewer, plus UI to build and reuse it. Works with any segmentation source — no dataset-specific code. I am using a custom file format for the cache file which is might not be ideal and to be discussed.

## What it does

- **`MeshCache`** (`~/.mobie/mesh-cache/<segmentation>-sm<k>-<spacing>um.mel`): binary per-segmentation cache of smoothed meshes. Memory-bounded: meshes are appended to the file whenever the in-memory buffer exceeds a threshold, and an on-disk index (label → offset) allows demand reads — so very large segment sets (e.g. tens of thousands of cells) can be pre-rendered without holding everything in RAM. Existing `.mel` format is versioned; nothing else changes on disk.
- **Pre-rendering** (`SegmentVolumeViewer.preRenderSegments`): computes + smooths meshes in parallel (shared thread pool), reports progress in the Fiji Log window and status bar (~10 s cadence), persists to disk, and continues gracefully past failures.
- **Resolution policy when no `resolution3d` is declared**: the finest resolution already present in the cache is used automatically. If no cache exists, a Log-window message explains that meshes will be computed on demand and how to pre-cache via the table's **Misc ▸ "Cache segment meshes…"**.
- **UI**: the segmentation table's Misc menu offers "Cache segment meshes...", with a dialog listing the segmentation's actual mipmap resolutions (µm) plus a custom spacing; defaults to the finest cached resolution.
- **Robustness**: if a segment has no voxels at the chosen mipmap level, extraction retries finer levels down to level 0; segments with no voxels at any level (e.g. table rows whose label is absent) are skipped and reported once, not as per-segment errors. Mesh-creation failures now keep their underlying cause.
- **Configurable cache location**: resolved by precedence — `-Dmobie.meshCacheDir` system property → `ij.Prefs` key `MoBIE.meshCacheDir` → portable default `<user.home>/.mobie/mesh-cache` (File-based, Windows-safe).

## Files

- New: `MeshCache`, `SegmentMeshCacher`, `MeshCacheTest`
- Modified: `SegmentVolumeViewer`, `MeshCreator`, `ViewManager`, `TableView`, `MoBIEHelper`, `UserInterfaceHelper`

## Validation

- `MeshCacheTest` (roundtrip, append across instances, auto-persist at small thresholds, overwrite of a cached label): green.
- Interactive runs validated in Fiji (Windows) on PlatyBrowser segmentation datasets (nuclei/cells/combined traces): caching at several resolutions, cache reuse on re-open, and the Log-window messages.

### DIFF
--- a/src/main/java/org/embl/mobie/lib/table/TableView.java
+++ b/src/main/java/org/embl/mobie/lib/table/TableView.java
@@ -38,6 +38,7 @@ import bdv.viewer.Source;
 import org.embl.mobie.lib.data.DataStore;
 import org.embl.mobie.lib.util.MoBIEHelper;
 import org.embl.mobie.lib.volume.MeshCache;
+import org.embl.mobie.lib.volume.SegmentMeshCacher;
 import org.embl.mobie.lib.image.Image;
 import org.embl.mobie.lib.image.NumericAnnotationImage;
 import org.embl.mobie.lib.serialize.View;
@@ -339,29 +340,7 @@ public class TableView< A extends Annotation > implements SelectionListener< A >
 		try
 		{
 			final SegmentationDisplay segmentationDisplay = ( SegmentationDisplay ) display;
-			if ( segmentationDisplay.segmentVolumeViewer == null )
-			{
-				IJ.showMessage( "The 3D segment viewer is not initialised for this table." );
-				return;
-			}
-
-			segmentationDisplay.segmentVolumeViewer.setVoxelSpacing( new double[]{ spacing, spacing, spacing } );
-			segmentationDisplay.segmentVolumeViewer.configureMeshCache( display.getName(), MoBIEHelper.getMeshCacheDir() );
-
-			Collection segments = tableModel.annotations();
-			if ( segments.isEmpty() )
-				segments = selectionModel.getSelected();
-			if ( segments.isEmpty() )
-			{
-				IJ.showMessage( "No segments to cache." );
-				return;
-			}
-
-			final int cachedBefore = segmentationDisplay.segmentVolumeViewer.getMeshCache().size();
-			IJ.showStatus( "Caching " + segments.size() + " segment meshes at " + formatSpacing( spacing ) + " um ..." );
-			segmentationDisplay.segmentVolumeViewer.preRenderSegments( segments );
-			final int newlyCached = segmentationDisplay.segmentVolumeViewer.getMeshCache().size() - cachedBefore;
-			IJ.showStatus( "" );
+			final int newlyCached = SegmentMeshCacher.cacheSegmentsAt( segmentationDisplay, spacing );
 			IJ.showMessage( "Done. Cached " + newlyCached + " segment meshes at " + formatSpacing( spacing ) + " um.\n\nCache: " + MoBIEHelper.getMeshCacheDir() );
 		}
 		catch ( Exception e )
--- a/src/main/java/org/embl/mobie/lib/table/TableView.java
+++ b/src/main/java/org/embl/mobie/lib/table/TableView.java
@@ -34,10 +34,13 @@ import net.imglib2.type.numeric.real.DoubleType;
 import org.embl.mobie.MoBIE;
 import org.embl.mobie.io.util.IOHelper;
 import org.embl.mobie.lib.annotation.AnnotatedRegion;
+import bdv.viewer.Source;
 import org.embl.mobie.lib.data.DataStore;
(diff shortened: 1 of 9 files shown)


cyrilcros:
I have tested this roughly, but I have not closely checked the code. This is more of a draft proposal.

The problem we have is that we may have a lot of segmented items to plot 500 traces / 10k nuclei / 11k cells.  The MoBIE viewer addresses those nicely by rendering each one in 3D one at a time, using a choice of resolution based on the size of the rendered object (so small nuclei are high resolution, but a nuclei plus trace have worse resolution because they are 'bigger'). The other issue is that this is slow and recomputed each time.

Instead, we are offering the possibility to load a segmentation table, go to `Misc /  Cache segment meshes…`, select a resolution and prerender everything to be at that resolution or better (if you render everything at 0.2 micron but an object is no bigger than 0.1, you would have a problem unless you adapted). Some Platy nuclei are 20 pixels only, they are segmentation leftovers.

The mesh is saved as `.mel` file kept locally, where each segmentation instance has a rendering that can be loaded independently. You can select 1000 nuclei and render them all much faster once this file exists, they still show up one at a time but there is no re-rendering.

This makes it possible to load large groups of segmented items in a reasonable time. The previous method was to load nuclei 1-1000, wait 30min, load 1001-2000, wait 30 mins etc...

You could try it by:

1. Opening Platybrowser
2. Opening the nuclei segmentation
3. Running `Misc /  Cache segment meshes…` from the table UI
4. Trying to render large subsets of nuclei

tischi:
Dear @cyrilcros this looks great! Thank you for the work! 

Please note that I changed the main branch now to the new version that I ship with Fiji-latest. Please merge the updated main branch into your mesh-cache feature branch.

And I see there is a merge conflict. This may be because of my change of the main branch. Could you please look into that?

cyrilcros:
Thanks @tischi I will try rebasing next week. I am also not sure this is really finished, it is more of a working draft.

The issue I have with that feature is that I was using a custom file format to store meshes, where each segmentation has an ID pointing to a mesh. That may be suboptimal.

For our more specialized Platybrowser stuff:

- /g/arendt/Cyril/cells-sm5-0_32um.mel had cells and /g/arendt/Cyril/nuclei-sm5-0_16um.mel had nuclei if you want to check by copying locally or symlinking. The render is faster but still imperfect. By default nothing is selected so the 3D view is black.

- There seems to have been some dynamic rendering based on the bounding box size before, which makes sense. I have noticed the cell segmentation has a lot of small scale bits and bobs that make no sense and caused bugs if I tried to render them if I do not fall back to the dynamic process. There is a pixel count somewhere for the cells, I believe, which sometimes ended as a single digit. Nobody wants to redo it, but I think no one did a clear size based filtering.

- I have been looking into error proofing the process, that is if you render something buggy you don't crash but report it and skip it, and explain what happened.

- The other issue I have not resolved super clearly is how to compromise on the rendering resolution for the traces.  Physically they take a large space, so by default we assigned them a low resolution.

tischi:
Dear @ekatrukha,

This is a pull request to MoBIE to cache meshes on disk. 

Do you maybe have some suggestions for how to do that properly? For example, what would be an appropriate file format?

ekatrukha:
Hello @tischi and @cyrilcros,

for the file format, the current implementation of binary dump I guess is ok, 
but there is no interoperability with other tools.
If you want that, my first guess would be [glTF](https://en.wikipedia.org/wiki/GlTF), kind of "industry standard",
it allows to store multiple meshes per file, can be compressed, mesh data as buffers already   + can be opened in Blender, BVB, etc. 
Plus, there is a pretty good Java library [JglTF](https://github.com/javagl/JglTF) for working with it, including reading/writing (I use it in BVB).

I would not go with PLY (one mesh per file) or Alembic (I do not know of convenient libraries).

Hope it helps!

Cheers,
Eugene

cyrilcros:
Thank you for the tip @ekatrukha

cyrilcros:
@tischi My update on top of your latest main failed because I am missing the POM for org.embl.mobie:mobie-io:jar:5.0.0-SNAPSHOT, you pushed something but the build failed at https://github.com/mobie/mobie-io/actions/runs/35063579726/job/104688906975 because of some Java 8 vs Java 11 mess

